PR #2, opened 2015-06-10
Remove brod module reference; it does not implement correct functions

dizzyd:
As it stands, the brod module does not actually implement the start function required when one does something like:

application:start(brod).


id:
Thanks for the patch!

PR #3, opened 2015-06-30
Rename start_* to start_link_* in brod's api

complex64:
Let the API reflect the semantics of starting a producer and consumer process: The OTP libraries offer both `start` and `start_link`, let us mimic its interface and follow the principle of least surprise.


id:
Good stuff.

Can we still leave the old functions, make them to call the new ones and mark as deprecated? We'll remove them a bit later.

complex64:
@id, that'd be the best - true. I added a commit on top including your proposed change :)

id:
Thank you!

PR #8, opened 2015-12-03
refactor producer

id:
Work in progress.

TODO:
- refactor consumer
- fix tests

### DIFF
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -24,10 +24,7 @@
 
 %% Producer API
 -export([ start_link_producer/1
-        , start_link_producer/3
-        , start_link_producer/4
-        , start_producer/1
-        , start_producer/3
+        , start_link_producer/2
         , stop_producer/1
         , produce/3
         , produce/4
@@ -71,39 +68,18 @@
 %%%_* Types --------------------------------------------------------------------
 -type host() :: {string(), integer()}.
 
-%%%_* Macros -------------------------------------------------------------------
--define(DEFAULT_ACKS,            1). % default required acks
--define(DEFAULT_ACK_TIMEOUT,  1000). % default broker ack timeout
-
 %%%_* API ----------------------------------------------------------------------
-%% @deprecated
-%% @equiv start_link_producer(Hosts)
--spec start_producer([host()]) -> {ok, pid()} | {error, any()}.
-start_producer(Hosts) ->
-  start_link_producer(Hosts).
-
-%% @deprecated
-%% @equiv start_link_producer(Hosts, RequiredAcks, AckTimeout, <<"brod">>)
--spec start_producer([host()], integer(), integer()) ->
-                   {ok, pid()} | {error, any()}.
-start_producer(Hosts, RequiredAcks, AckTimeout) ->
-  start_link_producer(Hosts, RequiredAcks, AckTimeout, ?DEFAULT_CLIENT_ID).
-
-%% @equiv start_link_producer(Hosts, 1, 1000)
+%% @equiv start_link_producer(Hosts, [])
 -spec start_link_producer([host()]) -> {ok, pid()} | {error, any()}.
 start_link_producer(Hosts) ->
-  start_link_producer(Hosts, ?DEFAULT_ACKS, ?DEFAULT_ACK_TIMEOUT).
-
-%% @equiv start_link_producer(Hosts, RequiredAcks, AckTimeout, <<"brod">>)
--spec start_link_producer([host()], integer(), integer()) ->
-        {ok, pid()} | {error, any()}.
-start_link_producer(Hosts, RequiredAcks, AckTimeout) ->
-  start_link_producer(Hosts, RequiredAcks, AckTimeout, ?DEFAULT_CLIENT_ID).
+  start_link_producer(Hosts, []).
 
 %% @doc Start a process to publish messages to kafka.
 %%      Hosts:
-%%        list of "bootstrap" kafka nodes, {"hostname", 1234}
-%%      RequiredAcks:
+%%        list of "bootstrap" kafka nodes, {"hostname", 9092}
+%%      Options:
+%%        list of tuples {atom(), any()} where atom can be:
+%%      required_acks:
 %%        How many acknowledgements the kafka broker should receive
 %%        from the clustered replicas before responding to the
 %%        producer.
@@ -116,7 +92,8 @@ start_link_producer(Hosts, RequiredAcks, AckTimeout) ->
 %%        number > 1 the broker will block waiting for this number of
 %%        acknowledgements to occur (but the broker will never wait
 %%        for more acknowledgements than there are in-sync replicas).
-%%      AckTimeout:
+%%        Default is -1.
+%%      ack_timeout:
 %%        Maximum time in milliseconds the broker can await the
 %%        receipt of the number of acknowledgements in
 %%        RequiredAcks. The timeout is not an exact limit on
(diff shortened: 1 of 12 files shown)


src/brod_utils.erl
```diff
@@ -27,6 +27,8 @@
         , get_metadata/1
         , get_metadata/2
         , try_connect/1
+        , now_microsec/0
```
zmstone: no caller found, remove ?

src/brod_producer_sup.erl
```diff
@@ -0,0 +1,61 @@
...
+        ]).
+
+%% Supervisor callbacks
+-export([init/1]).
+
+-define(SERVER, ?MODULE).
+
+%%%_* API functions ============================================================
+start_link() ->
+  supervisor:start_link({local, ?SERVER}, ?MODULE, []).
```
zmstone: {local, ?SERVER} implies only one such supervisor is allowed per-node.

src/brod_producer.erl
```diff
@@ -56,299 +50,286 @@
...
+init([Hosts, Options, Debug]) ->
   erlang:process_flag(trap_exit, true),
-  {ok, #state{ hosts       = Hosts
-             , debug       = Debug
-             , acks        = RequiredAcks
-             , ack_timeout = AckTimeout
-             , client_id   = ensure_binary(ClientId)
+  MaxLeaderQueueLen = proplists:get_value(max_leader_queue_len, Options,
+                                          ?DEFAULT_MAX_LEADER_QUEUE_LEN),
+  ets:new(?T_LEADERS, [protected, named_table]),
```
zmstone: using named table implies only one is allowed per-node

src/brod_app.erl
```diff
@@ -0,0 +1,56 @@
...
+%%        ]
+%%      }
+%%    ].
+%% @see brod:start_link_producer/2 for more info about all producer args
+%%
+%% @end
+start() ->
+  _ = application:load(brod), %% ensure loaded
+  %% if no producer is configured in app env, start the supervisor empty
+  PermanentProducers = application:get_env(brod, producers, _Default = []),
```
id: I think it's better to move this to brod_sup:init/1.
Then you don't need to ensure the app is loaded.

zmstone: start/0 is called by start/2 the application behaviour callback, application should have been loaded in normal cases (when started by application controller).
it is called here because we don't necessarily start the supervisor by application controller when testing.

dszoboszlay: I think the relation between `start/0` and `start/2` are usually the other way around:
- `start/2` is the application controller's callback function, and starts the supervision tree on its own. It can assume the application is loaded of course, so it can access the env variables.
- `start/0` is a convenience wrapper, and calls `application:ensure_[all_]started/1`.

The problem with this code is that `start/0` will start the supervision tree linked to the calling process (not what you'd expect), without starting the application (so subsequent attempts to start `brod` as an application will fail).

id: Right, we need to fix this.

dszoboszlay: On it.

src/brod_sup.erl
```diff
@@ -0,0 +1,73 @@
...
+  {ProducerId, Opts} =
+    case proplists:get_value(producer_id, Args) of
+      undefined ->
+        {ProducerId0, [{producer_id, ProducerId0} | Args]};
+      ID when is_atom(ID) ->
+        {ID, Args}
+    end,
+  { _Id       = ProducerId
+  , _Start    = {brod, start_link_producer, [KafkaHosts, Opts]}
+  , _Restart  = {permanent, RestartDelay}
```
dszoboszlay, line 65: What's the point of the restart delay (and therefore using the `brod_supervisor` instead of the standard `supervisor`) if the restart strategy is `{one_for_one, 0, 1}`, e.g. no restarts are allowed at all?

id, line 65: brod_supervisor is actually https://github.com/rabbitmq/rabbitmq-common/blob/master/src/supervisor2.erl.
RestartDelay is a cooldown period. When regular supervisor would crash, this one will try to restart a child after RestartDelay instead. And it will keep doing this infinitely.

zmstone:
i'm rewriting quite heavily, please hold on.

dszoboszlay:
@zmstone I won't change anything since you're working on this code, but will add a couple of comments on problems I see, if you don't mind. I can write fixes if you wish.

src/brod.erl
```diff
@@ -127,12 +111,17 @@ start_link_producer(Hosts, RequiredAcks, AckTimeout) ->
...
+%%        How many requests sent to a kafka broker can stay
+%%        unacknowledged by a broker before blocking producing thread.
+%%     max_requests_in_flight (optional, default = 5):
+%%        Worker process will be sending produce reqeusts to kafka
+%%        without waiting for acknowledgements of a previous one.
+%%        This is to control how many such requests can be sent.
+-spec start_link_producer([host()], proplists:proplist()) ->
+                             {ok, pid()} | {error, any()}.
+start_link_producer(Hosts, Options) ->
+  brod_producer:start_link(Hosts, Options).
```
dszoboszlay, line 124: Wouldn't it be nicer to start the producers under the `brod_sup` instead of linking them to the caller process?

src/brod_sup.erl
```diff
@@ -0,0 +1,73 @@
...
+
+%% @doc Start supervisor of permanent producers.
+%% @see brod:start_link_producer/2 for more info about all producer args
+%% @end
+-spec start_link([proplists:proplist()]) -> {ok, pid()}.
+start_link(PermanentProducers) ->
+  brod_supervisor:start_link({local, ?MODULE}, ?MODULE, PermanentProducers).
+
+init(PermanentProducers) ->
+  Children = [permanent_producer(P) || P <- PermanentProducers],
```
dszoboszlay, line 39: To interact with the producers you need to have their pids. We need an API to get the pids of the permanent producers from the supervisor.

src/brod_producer_sup.erl
```diff
@@ -0,0 +1,59 @@
...
+%%%_* API functions ============================================================
+start_link() ->
+  supervisor:start_link(?MODULE, []).
+
+start_worker(SupPid, Args) ->
+  supervisor:start_child(SupPid, Args).
+
+%%%_* Supervisor callbacks =====================================================
+init([]) ->
+  Spec = worker(brod_producer_worker, 5000, permanent),
```
dszoboszlay, line 43: This is definitely not OK. There's no `start_link/0` in `brod_producer_worker`. Most probably this supervisor should only have dynamically added children.

src/brod_producer.erl
```diff
@@ -56,300 +50,295 @@
...
-             , acks        = RequiredAcks
-             , ack_timeout = AckTimeout
-             , client_id   = ensure_binary(ClientId)
+  MaxLeaderQueueLen = proplists:get_value(max_leader_queue_len, Options,
+                                          ?DEFAULT_MAX_LEADER_QUEUE_LEN),
+  TabLeaders = ets:new(t_leaders, [protected]),
+  ets:new(?T_REQUESTS, [protected, named_table, ordered_set,
+                        {keypos, #request.ref}]),
+  {ok, SupPid} = brod_producer_sup:start_link(),
+  {ok, Brokers} = refresh_metadata(Hosts, TabLeaders),
```
dszoboszlay, line 123: Do not fetch the metadata or interact with Kafka in any way from the `init` function! If Kafka cannot be reached, this could crash your `init`. The supervisor will not restart a child who's `init` crashed, but terminate instead.

So a single network problem may prevent your entire application, and maybe the entire node from starting up.

zmstone:
thank you @dszoboszlay. but your are now commenting on the code that is going to be deleted from my new branch. will push quite soon.

src/brod_producer.erl
zmstone, line 123: noted. thanks

dszoboszlay:
OK, I will wait for the new branch to come out.

zmstone:
@dszoboszlay #11

zmstone:
close?

id:
@dszoboszlay are you ok if I close this one?

dszoboszlay:
Yes, of course.

id:
--> #11

PR #9, opened 2015-12-07
Remove timeouts from gen_server calls

dszoboszlay:
Both the consumer and the producer may attempt to connect to several Kafka nodes when handling requests. In case of network hiccups or Kafka nodes being unavailable such operations may take many seconds.
Therefore it's better not to use 5 second timeouts in calls to these processes.

### DIFF
--- a/src/brod.app.src
+++ b/src/brod.app.src
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Kafka client"},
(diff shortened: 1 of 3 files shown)


id:
Please bump version in src/brod.app.src.

dszoboszlay:
Done. (The old version was 1.5.0, but I bumped it to 1.5.2, because the last tag said 1.5.1.)

id:
:+1:

PR #11, opened 2015-12-08
Per partition producer

zmstone:
ready to merge


src/brod_producer.erl
```diff
@@ -48,459 +35,84 @@
...
-init([Hosts, RequiredAcks, AckTimeout, ClientId, Debug]) ->
+init({ClientId, Topic, Config}) ->
   erlang:process_flag(trap_exit, true),
-  {ok, #state{ hosts       = Hosts
-             , debug       = Debug
-             , acks        = RequiredAcks
-             , ack_timeout = AckTimeout
-             , client_id   = ensure_binary(ClientId)
-             }}.
+  {ok, Metadata} = brod_client:get_metadata(ClientId, Topic),
```
id, line 161: send a message to self() to do this as the first thing after init

src/brod_sup.erl
```diff
@@ -0,0 +1,88 @@
...
+%% TODO: add permanent producers
+%% @end
+-spec start_link() -> {ok, pid()}.
+start_link() ->
+  brod_supervisor:start_link({local, ?MODULE}, ?MODULE, clients_sup).
+
+%% @doc Start client under brod_sup.
+-spec start_client(client_id(), client_config()) -> ok | {error, any()}.
+start_client(ClientId, Config) ->
+  case brod_supervisor:find_child(brod_sup, ClientId) of
```
id: Don't think it's a good idea to rely on uniqueness of ClientId and force users to invent ones.

src/brod_producer.erl
```diff
@@ -48,459 +35,84 @@
...
-  end;
-handle_info({'EXIT', Pid, Reason}, State) ->
-  Sockets = lists:keydelete(Pid, #socket.pid, State#state.sockets),
-  %% Shutdown because cluster has elected other leaders for
-  %% every topic/partition most likely.
-  %% Users will have to resend outstanding messages
-  {stop, {socket_down, Reason}, State#state{sockets = Sockets}};
-handle_info(_Info, State) ->
+handle_info({'EXIT', _Pid, _Reason}, #state{} = State) ->
+  %% TODO: handle partition producer restart
```
id: why not start them under a simple one for one supervisor and let that one to handle restarts?

src/brod_sup.erl
zmstone: why not, is it that hard to invent an atom ?

id: Because kafka does not require it. Why force users to invent artificial names just to be able to use the library?

zmstone: I'll add a brod_client:start_link/1 so the users can call start_link(Hosts) and keep tracking/using the pid.
And I'll remove bord_sup:start_client/2, supervising an anonymous process is pointless in this case.

dszoboszlay:
I read the code and it looks like a big step forward! Nice separation of concerns with the introduction of connection, topic producer and partition producer processes. I found some problems here and there also, but those are easier to fix than to explain, so my PRs will come soon.

There are a couple of things I'd suggest to change, but I'd like to hear you opinion too. If you like my ideas, I'll implement them too. If you say no, then let's discuss how to make all of us happy.

### Put all processes under supervision

So currently predefined clients and topic producers go into the app's supervision tree, but you can also start stuff with `brod:start_link_*` that will be linked to the caller process instead.

I think it would make more sense to put these dynamic clients and topics under the same supervision tree too. It would also simplify some things (e.g. terminating producers if the client is stopped).

### Introduce symbolic names for protocol constants

So currently you can request ack-s from all in sync brokers by using the `{required_acks, -1}` option, where -1 is a magic constant in the Kafka wire protocol. It would be more Erlangish to introduce atoms instead of the magic constants, so one could write `{required_acks, all_in_sync}` for example.

### Monitors as async references

Both the producer and the socket allow async operations returning first a reference, then one or more follow up messages including that reference once an ack/response is received. The client then shall wait for such follow up messages, but also monitor the process, because there will be no ack/response of course if the producer/socket dies.

This could be simplified a bit if the reference would indeed be an active monitor on the producer/socket process. The cost is that the monitor would have to be removed once the ack/response is delivered.

id:
@dszoboszlay very good thoughts. Agree with each item except the last one :)
Why do you think it will simplify things if clients would have to demonitor brod process every time they receive an ack?
Can't they just call erlang:monitor/2 once they get the pid and keep it in the state?

src/brod.erl
```diff
@@ -62,48 +72,59 @@
...
+
+%% @doc Start a client.
+%% A client should work with only one kafka cluster
+%% Many producers may share the same client
+%% @end
+-spec start_link_client([endpoint()]) -> {ok, pid()}.
+start_link_client(Hosts) ->
+  brod_client:start_link(Hosts).
+
+%5 @doc Start a client under supervisor brod_sup.
```
zmstone: outdated

src/brod_client.erl
```diff
@@ -0,0 +1,249 @@
...
+handle_info(start_topics_sup, #state{client_id = ClientId} = State) ->
+  %% NOTE: in case this is not a permanent (configured in sys.config)
+  %%       the supervisor will be started empty (without any children).
+  {ok, Pid} = brod_sup:start_link_topics_sup(ClientId),
+  {noreply, State#state{topics_sup = Pid}};
+handle_info({'EXIT', Pid, _Reason}, #state{ meta_sock = Pid
+                                          , endpoints = Endpoints
+                                          } = State) ->
+  NewPid = start_metadata_socket(Endpoints),
+  {nereply, State#state{meta_sock = NewPid}};
```
zmstone: typo

src/brod_int.hrl
```diff
@@ -42,8 +42,27 @@
 -define(EC_NOT_ENOUGH_REPLICAS_AFTER_APPEND,
         'NotEnoughReplicasAfterAppendException').                           % 20
 
+-define(MAX_CORR_ID, 2147483647). % 2^31 - 1
+
+%% {cluster_id(), topic(), partition()} => pid()
+-define(PARTITION_WORKER_TAB, t_partition_workers).
```
zmstone: outdated

src/brod_producer.erl
```diff
@@ -48,459 +37,224 @@
...
+-spec get_partition_producer(pid(), binary()) -> {ok, pid()} | {error, any()}.
+get_partition_producer(Pid, KafkaMsgKey) ->
+  case erlang:process_info(Pid, dictionary) of
+    {dictionary, ProccessDict} ->
+      case proplists:get_value(brod_producer_type, ProccessDict) of
+        brod_partition_producer ->
+          %% the pid itself is a partition worker
+          {ok, Pid};
+        ?MODULE ->
+          %% the pid is a topic worker, get it from
```
zmstone: complete the comment

src/brod_sup.erl
```diff
@@ -0,0 +1,197 @@
...
+  { _Id       = Partition
+  , _Start    = {brod_partition_producer, start_link,
+                 [Client, Topic, Partition, Config, TopicProducer]}
+  , _Restart  = {permanent, DelaySeconds}
+  , _Shutdown = 5000
+  , _Type     = worker
+  , _Module   = [brod_partition_producer]
+  }.
+
+%% @private Get configured delay seconds bfore clients restart.
```
zmstone: bfore-> before

zmstone:
@dszoboszlay reply to your '###'s

### Put all processes under supervision:

This is already the case for permanent producers.
I left the flexibility for higher level users to work with producers directly
  --- the reason why start_link_xxx exists.

### Introduce symbolic names for protocol constants

I love this idea, but perhaps keep supporting integers too.

### Monitors as async references

I'm not sure if I fully understand what you meant.
But the current implementation is already monitoring the pid
just the reference is not the monitor reference.

```
103 -spec produce(pid(), binary(), binary()) ->
104         {ok, brod_produce_call()} | {error, any()}.
105 produce(Pid, Key, Value) ->
106   Ref = make_ref(),
107   Call = ?BROD_PRODUCE_CALL(self(), Ref),
108   ok = gen_server:cast(Pid, {produce, Call, Key, Value}),
109   %% Wait for BROD_PRODUCE_REQ_BUFFERED
110   case sync_produce_requests(Pid, [?BROD_PRODUCE_REQ_BUFFERED(Call)]) of
111     ok              -> {ok, Call};
112     {error, Reason} -> {error, Reason}
113   end.

115 -spec sync_produce_requests(pid(), [brod_produce_reply()]) ->
116         ok | {error, Reason::any()}.
117 sync_produce_requests(ProducerPid, ExpectedReplies) ->
118   Mref = erlang:monitor(process, ProducerPid),
119   Result = sync_produce_requests_loop(Mref, ExpectedReplies),
120   erlang:demonitor(Mref, [flush]),
121   Result. 

387 -spec sync_produce_requests_loop(ProducerPidMonitorRef::reference(),
388                                  [brod_produce_reply()]) ->
389         ok | {error, Reason::any()}.
390 sync_produce_requests_loop(_Mref, []) -> ok;
391 sync_produce_requests_loop(Mref, [Reply | Replies]) ->
392   receive
393     {'DOWN', Mref, process, _Pid, Reason} ->
394       {error, {producer_down, Reason}};
395     Reply ->
396       sync_produce_requests_loop(Mref, Replies)
397   end.
```

dszoboszlay:
@zmstone 

### Monitors

That's true, the producer staff is OK as it is. The `brod_sock:send_sync/3` is not monitoring though, but that would be a local a fix only.

``` erlang
send_sync(Pid, Request, Timeout) ->
  {ok, CorrId} = send(Pid, Request),
  receive
    %% Pid should be monitored here!
    {msg, Pid, CorrId, Response} -> {ok, Response}
  after
    Timeout -> {error, timeout}
  end.
```

I still have the feeling that something is not completely OK with the ack monitoring though. And I think I found just the correct counterexample. Consider this client code:

``` erlang
Prod = brod:start_link_producer(Client, Topic, Config),
{ok, CallRef} = brod:produce(Prod, Msg),
%% do something else, then decide to wait on the ack for Msg
ok = brod:sync_produce_request(Prod, CallRef).
```

This will eventually end up calling `brod_partition_producer:sync_produce_requests/2`, which will monitor the `Prod` pid, that is, the `brod_producer`. But we in fact need to monitor the `brod_partition_producer` instead. So the problem is with the API that allows producing via the topic process, but requires knowing about the topic producer to discover lost messages.

Turning the `CallRef` into a monitor on its own right would be one solution to the problem. But maybe not the best.

### Supervision

What I'm saying is that you cannot start a (topic) producer without first starting a client. And since the client already supervises its permanent producers, it would make sense to add the dynamic ones under the same supervisor too.

This would simplify stopping the process tree: the supervisor takes care of killing the producers, they don't have to actively monitor their client process any more.

The process calling `brod:start_link_producer/3` may still want to link to (or monitor) the producer though, so it will know when it dies. But it's not strictly necessary: any attempt to use the dead producer would crash the caller anyway.

The same logic could be applied to the client processes and the application's top level supervisor. Arguably `application:stop(brod)` should terminate all clients, not just the permanent ones.

zmstone:
i hate it when i can't reply in-line.

```
%% do something else, then decide to wait on the ack for Msg
ok = brod:sync_produce_request(Prod, CallRef).
%% by zmstone: this CallRef is a monitor of the brod_partition_producer, not brod_producer.
%% see below code:
193 produce(Pid, Key, Value) ->
194   case brod_producer:get_partition_producer(Pid) of
195     {ok, PartitionProducerPid} ->
196       brod_partition_producer:produce(PartitionProducerPid, Key, Value);
197     {error, Reason} ->
198       {error, Reason}
199   end.
200 
i.e. brod_producer is just used to get the correct partition_producer pid.

```

zmstone:
ah, i was wrong, sync_produce_requests monitored a wrong pid.

zmstone:
[quote]
The process calling brod:start_link_producer/3 may still want to link to (or monitor) the producer though, so it will know when it dies. But it's not strictly necessary: any attempt to use the dead producer would crash the caller anyway.
[quote]
agree.
then the APIs will look like this: ??
brod_clinet:start_link(Hosts, [{TopicName, TopicConfig}]).
then:
brod_client:get_producer_pid(ClientPid, TopicName)

@id @dszoboszlay sounds good?

id:
Yes, looks good.

include/brod.hrl
```diff
@@ -35,9 +35,33 @@
...
+-type brod_produce_result() :: brod_produce_req_buffered
+                             | brod_produce_req_acked.
+
+-record(brod_produce_reply, { call_ref :: brod_call_ref()
+                            , result   :: brod_produce_result()
+                            }).
+
+-type brod_produce_reply() :: #brod_produce_reply{}.
+
+-type brod_partitionner() :: random
```
id: brod_partitioner

id:
I'm still thinking that brod users should be isolated from workers restarts. The library should accept 'produce' requests from a user process and buffer them up to a certain limit while a partition is unavailable.
Right now we return {error, {producer_down, Reason}} in this case.

src/brod_producer.erl
```diff
@@ -16,493 +16,282 @@
...
+                                 , timeout = AckTimeout
+                                 , data    = Data
+                                 },
+      brod_sock:send(SockPid, KafkaReq)
+    end,
+  Buffer = brod_producer_buffer:new(BufferLimit, OnWireLimit,
+                                    MsgSetBytes, SendFun),
+  %% 5. Update state.
+  NewState = State#state{ sock_pid = SockPid
+                        , buffer   = Buffer
```
zmstone: update config

src/brod_producer.erl
```diff
@@ -16,493 +16,282 @@
...
+                                       ?DEFAULT_PARITION_ONWIRE_LIMIT,
+                                       Config1),
+  {MsgSetBytes, Config}  = take_config(message_set_bytes_limit,
+                                       ?DEFAULT_MESSAGE_SET_BYTES_LIMIT,
+                                       Config2),
+  RequiredAcks = get_required_acks(Config),
+  AckTimeout = get_ack_timeout(Config),
+  SendFun =
+    fun(KafkaKvList) ->
+      %% TODO: change to plain kv-list ?
```
zmstone: @id what do you think ?

id: We should only have one partition here, right?

zmstone: yes, if we stick to per-partition producer, all the requests are always sent per-partition, dict is kind of a cumbersome data structure since a plain list is enough.

id: yes, sure

zmstone:
closing this one, see #15

PR #12, opened 2015-12-14
rename + some patches

id:
It's a bit confusing with 'producer' and 'producers' and supervisor names.
Renamed:
brod_producer -> brod_partition_producer
brod_producers -> brod_producer

Also fixed few issues.


zmstone:
i'm resolving the conflicts.

PR #13, opened 2015-12-15
Simplify supervision tree for producer

id:


src/brod_client.erl
```diff
@@ -156,6 +156,9 @@ get_producer(ClientId, Topic, Partition) when is_atom(ClientId) ->
     {error, client_down}
   end.
 
+find_producer(ClientId, Topic, Partition) ->
```
zmstone, line 162: spec

zmstone, line 162: and export it since you moved it up the the API section

src/brod_producers_sup.erl
```diff
@@ -0,0 +1,139 @@
...
+
+producers_sup_spec(ClientId, TopicName, Config0) ->
+  DelaySecs = proplists:get_value(restart_delay_seconds, Config0,
+                                  ?DEFAULT_SUP2_RESTART_DELAY),
+  Config    = proplists:delete(restart_delay_seconds, Config0),
+  Args      = [?MODULE, {?SUP2, ClientId, TopicName, Config}],
+  { _Id       = TopicName
+  , _Start    = {brod_supervisor, start_link, Args}
+  , _Restart  = {permanent, DelaySecs}
+  , _Shutdown = infinity
```
zmstone, line 112: why infinity?

src/brod_supervisor.erl
```diff
@@ -1382,9 +1319,93 @@ init_state1(SupName, {Strategy, MaxIntensity, Period}, Mod, Args) ->
     period = Period,
     module = Mod,
     args = Args}};
-init_state1(_SupName, Type, _, _) ->
+init_state(_SupName, Type, _, _) ->
     {invalid_type, Type}.
 
+init_children(State, StartSpec) ->
```
zmstone: could you move this section back to the old place ?
this would make it more difficult to re-base or merge newer releases from otp.

src/brod_producers_sup.erl
```diff
@@ -0,0 +1,139 @@
...
+%%%_* APIs =====================================================================
+
+%% @doc Start a root producers supervisor,
+%%      per-topic supervisors and per-partition producer workers.
+%%      The config is passed down to the producers.
+%% @see brod_producer:start_link/4 for more details.
+%% @end
+-spec start_link(client_id(), [{topic(), producer_config()}]) -> {ok, pid()}.
+start_link(ClientId, Producers) ->
+  brod_supervisor:start_link({local, ?SUP}, ?MODULE,
```
zmstone: should not start registered.

src/brod_client.erl
```diff
@@ -186,8 +189,9 @@ handle_info({'EXIT', Pid, Reason}, State) ->
 handle_info(_Info, State) ->
   {noreply, State}.
 
-handle_call({get_topic_worker, Topic}, _From, State) ->
-  Result = do_get_topic_worker(State, Topic),
+handle_call({find_producer, Topic, Partition}, _From, State) ->
+  SupPid = State#state.producers_sup,
+  Result = brod_producers_sup:find_producer(SupPid, Topic, Partition),
```
zmstone: no, you can't do this in a handle_call, causing dead lock. (same reason why i implemented get_topic_worker before).
here you should return the second-level supervisor pid as quick as possible and then call brod_supervisor:find_child(Partition) in the caller process instead.

Reason:
brod_supervisor:find_child is a gen_server:call to SUP2, but in SUP2's post_init handle_info, it'll make a call to brod_client to get metadata, brod_client might be currently blocked by this find_producer call.

src/brod_producers_sup.erl
id, line 112: Recommended shutdown timeout for supervisors, they need to shut down children first.

PR #17, opened 2015-12-26
refactor consumer

id:
App is working but work is still in progress. Need to update app api and add tests.

Please leave your comments/complaints.


src/brod_client.erl
```diff
@@ -240,19 +288,22 @@ terminate(Reason, #state{ client_id     = ClientId
...
+      error_logger:warning_msg("~p [~p] ~p is terminating, reason: ~p~n",
+                               [?MODULE, self(), ClientId, Reason])
+  end,
+  %% stop producers and consumers first because they are monitoring socket pids
+  brod_utils:shutdown_pid(ProducersSup),
+  brod_utils:shutdown_pid(ConsumersSup),
+  lists:foreach(
+    fun(#sock{sock_pid = Pid}) ->
+        brod_utils:shutdown_pid(Pid)
+    end, [MetaSock | Sockets]).
```
zmstone, line 319: MetaSock might be ?undef

src/brod_client.erl
```diff
@@ -66,30 +68,60 @@
...
+  gen_server:start_link({local, ClientId}, ?MODULE, Args, []).
 
 -spec stop(client()) -> ok.
 stop(Client) ->
   gen_server:call(Client, stop, infinity).
 
+%% @doc Returns brod_sock connected to a broker which is a leader
+%% for given topic and partition.
+-spec get_leader(client_id(), topic(), partition()) -> {ok, pid()}.
+get_leader(ClientId, Topic, Partition) ->
```
zmstone: rename it to: get_leader_connection and remove get_connection/3 ?

src/brod_client.erl
```diff
@@ -66,30 +68,60 @@
...
+                     , leader_id  = LeaderId} =
+    lists:keyfind(Partition, #partition_metadata.id, Partitions),
+  brod_kafka:is_error(PartitionEC) andalso erlang:throw(PartitionEC),
+  LeaderId >= 0 orelse erlang:throw({no_leader, {ClientId, Topic, Partition}}),
+  #broker_metadata{host = Host, port = Port} =
+    lists:keyfind(LeaderId, #broker_metadata.node_id, Brokers),
+
+  case brod_client:get_connection(ClientId, Host, Port) of
+    {ok, Pid}      -> {ok, Pid};
+    {error, Error} -> exit({no_connection, Error})
```
zmstone: why exit here?
perhaps replace the 'throw's above to exit too ?

src/brod_client.erl
```diff
@@ -206,14 +248,18 @@ handle_info({'EXIT', Pid, Reason},
                         [ClientId, Host, Port, Reason]),
   NewSock = start_metadata_socket(ClientId, Endpoints),
   {noreply, State#state{meta_sock = NewSock}};
-handle_info({'EXIT', Pid, Reason}, State) ->
+handle_info({'EXIT', Pid, Reason}, #state{meta_sock = Pid} = State) ->
```
zmstone: meta_sock is a #sock{} record.
and, it's handled in handle_info clause line 240

src/brod_consumer.erl
```diff
@@ -48,238 +42,209 @@
...
         , code_change/3
-        , format_status/2
         ]).
 
 -include("brod.hrl").
 -include("brod_int.hrl").
 
--record(state, { hosts         :: [{string(), integer()}]
-               , socket        :: #socket{}
+-type consumer_option() :: offset
```
zmstone: we have used the name 'config' for clients and producer.
i.e. client_config and producer_config.
perhaps consumer_config here ?

sys.config.example
```diff
@@ -7,7 +7,7 @@
              , [ {restart_delay_seconds, 10}]
              }
            , { producers
-             , [ { <<"brod-test-topic-1">>
+             , [ { <<"brod">>
```
zmstone: used "topic" in the name to make it clear that it should be a topic name here.
if rename, perhaps add a comment behind <<"brod">> ?
e.g. <<"brod">> %% topic name

src/brod_consumer.erl
```diff
@@ -48,238 +42,209 @@
...
+      %% TODO: do something smarter
+      %% 1) handle what we can handle
+      %% 2) call CbMod:handle_error for errors we can not handle
+      %%    it can for example bump max_bytes option or skip a message
+      %% 3) stop if CbMod:handle_error is not defined
+      {stop, {broker_error, ErrorCode}, State0};
+    false ->
+      case Messages of
+        [] ->
+          erlang:send_after(State0#state.sleep_timeout, self(), ?SEND_FETCH_REQUEST),
```
zmstone: self() ! ?SEND_FETCH_REQUEST directly if sleep_timeout is zero.
so it doesn't have to schedule a timer

src/brod_consumer.erl
id: consumer_config is already defined in brod_int.hrl, it's proplists:proplist().

zmstone:
perhaps add brod:fetch(Client, Topic, Partition, MaxWaitTime, MinBytes, MaxBytes), which should make a call to partition worker?

src/brod_consumer.erl
zmstone: ok, new question, why would a user put those options in config ?

zmstone:
now i'm not quite sure about callback modules, or perhaps callback module is not just enough.
as I can think of now, it's perhaps more flexible to provide an async fetch API, which should work like:

```
brod:fetch(Client, Topic, Partition, MaxWaitTime, MinBytes, MaxBytes) ->
  {ok, ConsumerPid} = brod_client:get_consumer(Client, Topic, Partition),
  CallRef = {make_ref(), _RealConsumerPid = self()},
  brod_consumer:async_fetch(CallRef, ConsumerPid, MaxWaitTime, MinBytes, MaxBytes)
```

the CallRef should be kept in consumer's looping state ,
the consumer should deliver the message(s) to RealConsumerPid when fetch response is received.

AND, if configured, consumer process should expect RealConsumerPid to ACK the fetch response before removing a {CallRef, offset()} association from looping state --- this is to be used for offset commit too.

This way, it is completely up to the RealConsumerPid how many messages they'd like to fetch in parallel before stop sending fetch requests.

src/brod_consumer.erl
id: The only option which I know exactly will be used is 'offset' – when users will use their own storage for offset, in the 'init_consumer' callback they will query offset for the given topic/partition pair.
Other options can be used to override global options for specific topic/partitions.

zmstone: does this imply: always have to restart the consumer process to reset offset ?
for example, if a consumer wants to search for a offset <--> timestamp (in message key) correlation in order to consume messages starting from a specific timestamp, how should they implement the callback module ?

zmstone:
I'd prefer below APIs over callback module.
if there is really a need of callback module, it can be implemented on top of these APIs

```
%% Consumer API
-export([ async_fetch/5 %% (Client, Topic, Partition, BeginOffset, Options) -> ok. %% call the permanent consumer process to send fetch requests, caller should expect {brod_messages, MsgRef, [{Topic, Partition, Offset, Key, Value}]} where MsgRef is CorrId (which is perhaps discarded by direct users)
        , start_link_partition_streamer/4 %% (Client, Topic, Partition, BeginOffset, Options) -> {ok, Pid}. %% start_link a partition streamer process which accepts a [{prefetch_count, N}] in option and call aync_fetch behind the scean, caller should expect {brod_messages, MsgRef, [[Topic, Parttion, Offset, Key, Vlaue}]} where MsgRef is of type {PartitionStreamerPid, CorrId}.
        , start_link_group_streamer/4 %% (Client, GroupId, Options) -> {ok, Pid}. %% start_link a group streamer process which streams from one or more partition streamers and manages group heartbeats, membership changes etc. caller should expect {brod_messages, [[Topic, Parttion, Offset, Key, Vlaue}]} where MsgRef is of type {GroupStreamerPid, PartitionStreamerPid, CorrId}.
        , stream_ack/2 %% (MsgRef, Offset), this is called to delete the {CorrId, [Offset]} associations from streamers. they should continue if prefetch window size allows. also commit offset if it's an ack to group streamer, where MsgRef can be {PartitionStreamerPid, CorrId} or {GroupStreamerPid, PartitionStreamerPid, CorrId}.
```

id:
Don't understand what is the problem with callbacks modules. Just add -behaviour(brod_consumer) in e.g. your gen_server implementation and define 2 more functions.

With async_fetch there is very little logic left for a library module. Basically, a user can ask brod_client for connection pid and just call brod_sock:send(Pid, #fetch_request{}) with the same effect. For this to work we just need to move record definitions from src/brod_int.hrl to include/brod.hrl.

id:
group streamer: so you suggest to collect messages from many partitions in one process and then re-send to a user process? What is the benefit?

id:
I would say that these APIs can be easily implemented using behaviour module if users need them.

zmstone:
not a very good idea to let the higher level callers to call brod_sock:send.
with generic APIs, we maximize the flexibility for higher level users by pure erlang message sending, and by default it encourages async pre-fetching way of working.

If brod supports callback module by default, it is tied to one message set at a time, unless we add more configurable options like: prefetch_count, is_parallelizable_callback_evaluation etc.

and as you said, the init callback is only useful to init the offsets, the users may have to call another API to figure out the offset first before initializing the behaviour module.

as for group streamer, it is the nature of consumer group, it does not know how many partitions it will have to fetch from, and as a brod user, i don't want to manage those contexts.

zmstone:
partition streamer and group streamer are better to be implemented in one gen_server each to keep code logic simple.
implementing everything in one process will be bad for test/debug and maintenance.

id:
Of course, I'm talking about another concern.

If I understood you correctly, group streamer process will act like a hub for all messages from group partitions. It will then redeliver messages to a user process's Pid. And that process would have to implement a logic to process messages in parallel. Comparing to a callback we have 2 more middle-man processes and additional logic to be implemented by a user.

zmstone:
1. even if you implement callback for consumer group, it will be another gen_server, not the one that consumes from only one partition.
2. with start_link_xxx_streamer, it can talk to brod_sock directly, meaning there is no need to have a consumer under supervisor tree, so -1 middle-man, does this make you feel any better ? :)

if we see a lot of use cases where some dummy callbacks like below will have to be implemented for certain purposes:

```
handle_messages(Topic, Partition, HighWmOffset, Messages, #my_state{my_pid = PId}) ->
  Pid ! {Topic, Partition, HighWmOffset, Messages}.
```

it means a bad design.

It will perhaps make me feel better if it extends all gen_server callbacks, plus, force users to implement an additional handle_message callback

however, the init callback is really something tricky, same will be for commit offset.

Regarding message processing:
The caller process can process messages in parallel (e.g. pmap) or maybe not, it's up to them.
However now i realize this is something can be done (in one message-set level) in callback too, since the callback argument is MessageSet, it's up to the caller if they want to process messages in one message set in parallel or not. It is much better than what I thought it was before. I thought it was something like foreach(fun(message) -> CbMod:handle_message(OneMessage) end, MessageSet).

let's drop this, and focus on which style is more user-friendly.

id:
I'll try to summarize what I have in mind about possible use cases for consumers.

Use case #1. I want to read data from kafka, process it, and write to postgres with minimal efforts.
Solution: add brod in the release and implement brod_consumer behaviour. 
- In the init_consumer callback read the last offset from postgres.
- In the handle_messages callback transform a message to a sql query and send to a pool of epgsql worker processes.
- Add required information in config file and make sure brod app is started after epgsql pool.
- All processing is happening in the context of partition workers and already parallelized, don't need to do anything special.
  Downside: in the current implementation consumer process will be blocked by the callback, may be need to send next fetch request before calling callback (kind of prefetch count = 1).

Use case #2. I need to get data from kafka in my existing process (a gen_server implementation in existing application).
Solution: 
- start_link brod consumer process in the init function
- that process should be able to handle reconnects to cluster nodes without crash
- should be possible to configure 'prefetch count'
- flow control via 'acks'.

Am I correct? What do you think?

zmstone:
yes, sort of what was in my mind too. I have a few more things to add though.

use case 1:
proposed spec:

```
init_consumer(Topic, Partition) -> {ok, Options, CallbackState}.
```

Where Options is an minimal proplist with possible keys: begin_offset, max_bytes, prefetch_count.
Value for other options (e.g. sleep_time) are used in the order of: Options --(if not found)--> sys.config --(if not found) --> default values.

```
handle_messages(HwOffset, [{Key, Value}], CallbackState) ->
  {ok, NewCallbackState} | {ok, NewOptions, NewCallbackState}.
```

There is no need for Topic and Partition in handle_messages arguments, because it's a partition consumer, if the user need it, it can be kept in CallbackState.
By returning NewOptions, the user may switch to another begin_offset in case the last delivery was not they wanted.
And in order not to block the consumer process, handle_message should be evaluated in a spawned temp process. then fetch more if prefetch_count allows.

We'll need another behaviour for group consumer, which should be similar, but a bit different in Arg list and return value:

```
handle_messages(Topic, Partition, HwOffset, [{Key, Value}], CallbackState) ->
    {ok, NewCallbackState} | {ok, commit, NewCallbackState}.
```

id:
"There is no need for Topic and Partition in handle_messages arguments, because it's a partition consumer" – but users can have many partition consumers.

Passing #message{} record in handle_messages is very important, it contains offset and other message metadata. HighWmOffset is just a max offset on a partition.

"And in order not to block the consumer process, handle_message should be evaluated in a spawned temp process. then fetch more if prefetch_count allows." – and if handle_messages returns different begin_offset, then what?

src/brod_consumer.erl
```diff
@@ -48,265 +42,370 @@
...
-               , sleep_timeout :: integer()
+-type consumer_option() :: offset
+                         | min_bytes
+                         | max_bytes
+                         | max_wait_time
+                         | sleep_timeout
+                         | prefetch_count.
+
+%% behaviour definition
+-callback init_consumer( Topic     :: topic()
```
zmstone: add more docs?

btw, why don't return CbOptions (or NewCbOptions) as the second element in the returned tuple ?

src/brod_consumer.erl
```diff
@@ -48,265 +42,370 @@
...
-               , socket        :: #socket{}
-               , topic         :: binary()
-               , partition     :: integer()
-               , callback      :: callback_fun()
-               , offset        :: integer()
-               , max_wait_time :: integer()
-               , min_bytes     :: integer()
-               , max_bytes     :: integer()
-               , sleep_timeout :: integer()
+-type consumer_option() :: offset
```
zmstone: name it begin_offset ?

src/brod_kafka.erl
```diff
@@ -55,7 +55,7 @@ parse_stream(<<Size:32/?INT,
   Response = decode(ApiKey, Bin),
   parse_stream(Tail, [{CorrId, Response} | Acc], Requests);
 parse_stream(Bin, Acc, _Requests) ->
-  {Bin, Acc}.
+  {Bin, lists:reverse(Acc)}.
```
zmstone: this should be ported back to 1.0

src/brod_utils.erl
```diff
@@ -58,15 +60,20 @@ try_connect([{Host, Port} | Hosts], _) ->
...
+%% @doc Check terminate reason for a gen_server implementation
+is_normal_reason(normal)        -> true;
+is_normal_reason(shutdown)      -> true;
+is_normal_reason({shutdown, _}) -> true;
+is_normal_reason(_)             -> false.
+
+is_pid_alive(Pid) ->
+  is_pid(Pid) andalso is_process_alive(Pid).
+
+shutdown_pid(Pid) ->
```
zmstone, line 72: perhaps we should monitor, unlink, exit, and then wait for 'DOWN'.
or a batched version of it ( for shutting down a list of sockets) .

It tricky on how long should it wait for the 'DOWN' message though. and it is somehow affecting the shutdown timeouts in supervisor (if any).

sys.config.example
```diff
@@ -15,6 +15,22 @@
                  }
                ]
              }
+           , { consumers
+             , [ { <<"brod-test-topic-1">>
+                   , [ {topic_restart_delay_seconds, 10}
+                     , {partition_restart_delay_seconds, 2}
+                     %% no 'partitions' means 'all partitions'
+                     % , {partitions, [1,2,3]}
```
zmstone, line 23: or {partitions, []} in the example, and force it to be a mandatory config.

src/brod_consumer.erl
```diff
@@ -48,265 +42,370 @@
...
+      SleepTimeout = State0#state.sleep_timeout,
+      case Messages of
+        [] when SleepTimeout =:= 0 ->
+          {ok, NewCorrId} = send_fetch_request(State0),
+          {noreply, State0#state{corr_id = NewCorrId}};
+        [] when SleepTimeout > 0 ->
+          erlang:send_after(SleepTimeout, self(), ?SEND_FETCH_REQUEST),
+          {noreply, State0};
+        [_|_] ->
+          F = fun() ->
```
zmstone: I feel like using exit reason to pass result is a bit less readable comparing to straightforward message passing, and there is no need to monitor.

```
Ref = make_ref(),
ConsumerPid = self(),
spawn_link(fun() ->
  try
    Res = CbMod:handle_messages(...),
    ConsumerPid ! {handle_messages_result, Ref, ok, Res}
  catch C : E ->
    Stacktrace = erlang:get_stacktrace(),
    ConsumerPid ! {handle_messages_result, Ref, error, {C, E, Stacktrace}}
  end)
```

zmstone: and this may cause handle_messages evaluated in parallel which is in most cases a bad thing.

it should rather be designed as a fetch queue, and only the head of the queue get popped out for handle_messages evaluation. i.e. one at a time.

src/brod_consumer.erl
```diff
@@ -48,265 +42,370 @@
...
+               , socket_pid     :: pid()
+               , topic          :: binary()
+               , partition      :: integer()
+               , offset         :: integer()
+               , max_wait_time  :: integer()
+               , min_bytes      :: integer()
+               , max_bytes      :: integer()
+               , sleep_timeout  :: integer()
+               , prefetch_count :: integer()
+               , corr_id        :: corr_id()
```
zmstone: last_corr_id ?

src/brod_consumer.erl
id: Why is that a bad thing?

src/brod_consumer.erl
id: > why don't return CbOptions (or NewCbOptions) as the second element in the returned tuple ?

To keep the order of elements the same in both cases.

src/brod_consumer.erl
zmstone: because by providing this callback framework, we imply CbState is a state machine, almost all the users will assume the messages are delivered and handled in serial.

id: How do we treat prefetch_count then? Same as now (number of fetch requests sent) or as max number of messages?

> it should rather be designed as a fetch queue

How would you do this?

zmstone: we can: instead of put monitor reference in cb_pending, append a fun((CbState) -> NewCbState)) to the tail of the list.
then spawn a pid to evaluate the only head of the list, put back a reference() to the head to match the async evaluation result.

src/brod_consumer.erl
```diff
@@ -48,265 +42,394 @@
...
+                     , error_code = ErrorCode
+                     , high_wm_offset = HighWmOffset
+                     , last_offset = LastOffset
+                     , messages = Messages} = PM,
+  CbMod = State0#state.cb_mod,
+  CbState = State0#state.cb_state,
+  case brod_kafka:is_error(ErrorCode) of
+    true ->
+      Error = {ErrorCode, brod_kafka_errors:desc(ErrorCode)},
+      case CbMod:handle_error(Topic, Partition, Error, CbState) of
```
zmstone: there is still an issue, handle_error should also be lined up

src/brod_consumer.erl
id: Ok, got it. I misinterpreted your comment first and thought you're suggesting to have rather a 'handle_message' callback, i.e. call it for every message.

src/brod_client.erl
```diff
@@ -66,22 +68,55 @@
...
 
+%% @doc Get the connection to kafka broker which is a leader
+%% for given Topic/Partition.
+%% If there is no such connection established yet, try to establish a new one.
+%% If the connection is already established per request from another
+%% producer/consumer the same socket is returned.
+%% If the old connection was dead less than a configurable N seconds ago,
+%% {error, LastReason} is returned.
+%% @end
+-spec get_leader_connection(client_id(), topic(), partition()) -> {ok, pid()}.
```
zmstone: | {error, Reason}
we need dialyzer

src/brod_client.erl
```diff
@@ -154,38 +199,59 @@ get_producer(ClientId, Topic, Partition) when is_atom(ClientId) ->
...
-             , config    = Config
+  Tab = ets:new(?ETS(ClientId),
+                [named_table, protected, {read_concurrency, true}]),
+  self() ! {init, Producers, Consumers},
+  {ok, #state{ client_id     = ClientId
+             , endpoints     = Endpoints
+             , config        = Config
+             , producers_tab = Tab
+             }};
+init({Endpoints, Producers, Consumers, Config}) ->
```
zmstone, line 219: if producers are started here, get_producer will always return {error, client_down}.
part of the reason why i thought it's good to always force a unique client_id is to be able to use the name as the ets table, so the callers can get the producer pid from ets:lookup instead of going through a gen_server:call for each produce request.

id, line 219: Unique ClientId can be useful for producers but gives no benefit (only overhead) for consumers.

I tried to reduce dependency on ClientId as much as possible. If all I want is just start a consumer, why on Earth I must think about a unique name for it?

If users will decide to use brod_client:stat_link/4 instead of brod_client:start_link/5, or even brod:start_link_client, they should know what they are doing.

src/brod_client.erl
```diff
@@ -154,38 +199,59 @@ get_producer(ClientId, Topic, Partition) when is_atom(ClientId) ->
...
+init({ClientId, Endpoints, Producers, Consumers, Config}) ->
   erlang:process_flag(trap_exit, true),
-  ets:new(?ETS(ClientId), [named_table, protected, {read_concurrency, true}]),
-  self() ! {init, Producers},
-  {ok, #state{ client_id = ClientId
-             , endpoints = Endpoints
-             , config    = Config
+  Tab = ets:new(?ETS(ClientId),
+                [named_table, protected, {read_concurrency, true}]),
+  self() ! {init, Producers, Consumers},
```
zmstone, line 213: register(ClientId, self()).

id, line 213: If we are here, then start_link/5 was called and the process is already registered.

src/brod_client.erl
zmstone, line 219: registered name causes overhead ?

id, line 219: A simple test to show that this optimization does not contribute that much for performance.

I added one more field in brod_client #state{}, a dict to keep a map {T, P} -> ProducerPid, in addition to ets table. And replaced the first clause of brod_client:get_producer api which matches on pid to do gen_server:call instead and do dict:find.

Results are vary and this is the best one so far :), but the slowest gen_server:call was less than 150 microseconds for me:

```
13> timer:tc(fun() -> brod_client:get_producer(client_1, <<"brod">>, 0) end).
{44,{ok,<0.43.0>}}
14> timer:tc(fun() -> brod_client:get_producer(Pid, <<"brod">>, 0) end).
{55,{ok,<0.43.0>}}
```

id, line 219: "registered name causes overhead ?" – overhead as in a user have to think about additional things which they will not use.

zmstone, line 219: with ets is you don't have to line up the calls when you have many processes calling get_producers concurrently.
if this performance improvement is really not benefiting, pg2 would have been implemented with pure gen_server calls plus dict or sets, instead of storing members in ets.

src/brod_client.erl
zmstone, line 213: missed that one

src/brod_client.erl
id, line 219: True.

What I want to say is this part of brod is hardly a bottleneck. At least for the most of use cases.

id:
@dszoboszlay could you also take a look at the new consumers implementation? We appreciate your input.

id:
Docs and tests are still on me.

id:
@zmstone I think we can merge this one to 2.0-dev, wdyt?

zmstone:
#23 is based on this one. we should merge it in first.

id:
Don't see why it's a block. #23 will have the same commits, we do not rewrite history.

id:
I'll let the checks to complete first.

PR #20, opened 2016-01-10
Smarter connection management

### LINKED ISSUE #19: Producer leader caching
The way brod producer caches leaders leads to creating way too many sockets. 

The Producer creates a proplist of [{{Topic, Partition}, brod_sock Pid} | Leaders ]

For a fairly typical case of having say 10 topics with 15 partitions each you would create 150 socket connections. Even if you have a typical 3 node kafka this ends up being 50 connections to each kafka. 

We discovered this through testing brod producing to this setup
- single kafka
- 1 topic
- 11 partitions
- 10 poolboy workers each with a producer

After messing with our worker size we discovered that we can easily flood kafka with connections .  After adding some debug to brod we saw that we are creating 11 sockets per producer when 1 would do. 

Any thoughts?

id:
fixes #19

### DIFF
--- a/src/brod.app.src
+++ b/src/brod.app.src
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Kafka client"},
-  {vsn,"1.6.2"},
+  {vsn,"1.5.3"},
   {registered,[]},
   {applications,[kernel,stdlib]},
   {env,[]},
--- a/src/brod.app.src
+++ b/src/brod.app.src
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Kafka client"},
-  {vsn,"1.5.2"},
+  {vsn,"1.6.2"},
   {registered,[]},
   {applications,[kernel,stdlib]},
   {env,[]},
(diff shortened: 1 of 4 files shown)


src/brod.app.src
```diff
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Kafka client"},
-  {vsn,"1.5.2"},
```
zmstone, line 4: bug fix --> 1.5.3

src/brod_producer.erl
```diff
@@ -270,20 +270,22 @@ get_leader(Topic, Partition, State) ->
...
+      NewLeader = lists:map(fun({T, P}) -> {{T, P}, Pid} end, LeaderPartitions),
+      Leaders = lists:append(NewLeader, State#state.leaders),
       {ok, Pid, State#state{sockets = Sockets, leaders = Leaders}};
     {error, Reason} ->
       {error, Reason}
   end.
 
--spec fetch_leader_broker_metadata(topic(), partition(), #state{}) ->
+-spec fetch_metadata(topic(), partition(), #state{}) ->
         {ok, #broker_metadata{}} | {error, any()}.
```
zmstone: update spec

zmstone:
good job

zmstone:
merge ?

PR #21, opened 2016-01-16
New consumer api

id:
Implement offset commit/fetch api.
Tests later.

Works with Kafka 0.9.


zmstone:
turned out 0 changes after merge :)
@id it was only the message encode / decode added in this branch right ?
if so, delete this branch and start over for the new apis ?

id:
Interesting :)

Yes, sure.

PR #22, opened 2016-01-16
Subscribe to consume

zmstone:


src/brod.erl
```diff
@@ -212,16 +225,38 @@ sync_produce_request(CallRef) ->
...
-                         , topic()
-                         , consumer_config()) ->
-                             {ok, pid()} | {error, any()}.
-start_link_consumer(Subscriber, Endpoints, Topic, Options0) ->
-  Options = [ {cb_mod, brod_consumer_impl}
-            , {cb_args, Subscriber} | Options0],
-  Consumer = {Topic, Options},
-  brod_client:start_link(Endpoints, [], [Consumer], []).
+%% @doc Subscribe data stream from the given topic-partition.
+%% If {error, {producer_down, noproc} is returned, the caller should
```
id: producer_down?

id: Ok, it must be 'consumer_down'.
In which situation we get it?

src/brod_consumer.erl
```diff
@@ -163,124 +142,88 @@ handle_info(init_socket, #state{ cb_mod     = CbMod
...
-  {noreply, State#state{last_corr_id = CorrId}};
+handle_info({'DOWN', _MonitorRef, process, Pid, _Reason},
+            #state{subscriber = Pid} = State) ->
+  error_logger:info_msg("~p ~p subscriber ~p is down",
+                        [?MODULE, self(), Pid]),
+  NewState = reset_buffer(State#state{subscriber = undefined}),
+  {noreply, NewState};
 handle_info({'DOWN', _MonitorRef, process, Pid, Reason},
             #state{socket_pid = Pid} = State) ->
   {stop, {socket_down, Reason}, State};
```
id: Shouldn't stop here, need to keep retrying.

I'll submit a PR with producer refactoring to show what I mean soon.

src/brod.erl
```diff
@@ -212,16 +225,38 @@ sync_produce_request(CallRef) ->
...
-                         , [endpoint()]
-                         , topic()
-                         , consumer_config()) ->
-                             {ok, pid()} | {error, any()}.
-start_link_consumer(Subscriber, Endpoints, Topic, Options0) ->
-  Options = [ {cb_mod, brod_consumer_impl}
-            , {cb_args, Subscriber} | Options0],
-  Consumer = {Topic, Options},
-  brod_client:start_link(Endpoints, [], [Consumer], []).
+%% @doc Subscribe data stream from the given topic-partition.
```
id: need description on which messages users should expect

src/brod_client.erl
```diff
@@ -159,67 +184,36 @@ get_partitions(Client, Topic) ->
...
-    {error, client_down}
-  end.
+%% @doc Register self() as a partition consumer. The pid is registered in an ETS
+%% table, then the callers may lookup a consumer pid from the table ane make
+%% subscribe calls to the process directly.
+%% @end
+register_consumer(Client, Topic, Partition) ->
+  Consumer = self(),
+  Key = ?CONSUMER_KEY(Topic, Partition),
+  gen_server:cast(Client, {register, Key, Consumer}).
```
id, line 204: I think it should be gen_server:call for both producer and consumer

src/brod_consumer.erl
```diff
@@ -292,123 +235,130 @@ code_change(_OldVsn, State, _Extra) ->
...
+  end.
+
+handle_message_set(#kafka_message_set{messages = []}, State) ->
+  ok = maybe_delay_fetch_request(State),
+  {noreply, State};
+handle_message_set(#kafka_message_set{messages = Messages} = MsgSet,
+                   #state{ subscriber    = Subscriber
+                         , pending_acks  = PendingAcks
+                         } = State0) ->
+  ok = cast(Subscriber, MsgSet),
```
id: I think we can include Ref and self() in this message, so that users can 'ack' using info from the message itself.

src/brod.erl
zmstone: when restarting.

src/brod_client.erl
zmstone, line 204: why ?

id, line 204: How do you make sure if a producer/consumer process has registered otherwise?

zmstone, line 204: cast should always work (sooner or later)
unless client process restart --- but, if that's the case, all producers and consumers should be restarted too.

id, line 204: True.

src/brod_consumer.erl
zmstone: like: #kafka_message_set.producer_pid ?

id: Subscriber ! {Ref, Pid, #kafka_message_set{}}.

src/brod_consumer.erl
```diff
@@ -418,38 +395,63 @@ do_debug(Pid, Debug) ->
...
+                           , partition    = Partition
+                           } = State) ->
+  case fetch_valid_offset(SocketPid, BeginOffset, Topic, Partition) of
+    {ok, NewBeginOffset} ->
+      {ok, State#state{begin_offset = NewBeginOffset}};
+    {error, Reason} ->
+      {error, Reason}
+  end.
+
+fetch_valid_offset(SocketPid, InitialOffset, Topic, Partition) ->
```
zmstone, line 438: @id, why do we need this ?

id, line 438: Kafka does not understand -1 and -2 in fetch requests, you first need to resolve this to a valid number via offset request.

zmstone, line 438: then perhaps we should be specific to only resolve just -1 and -2?
it would simplify the code logic and save some unnecessary round trips.

id, line 438: Will not handle all the cases.
Some old offsets that consumers saved might not work as well. This request will return something valid in this case also.

Round trips are only on restart/reconnect, why do you think it's a problem?

zmstone, line 438: i found this issue when testing socket restart:
if the 'next offset to be' is N, number N will be in the consumer looping state as the next 'begin_offset'.
but if you try to resolve this offset, kafka will return [] in the result.

the current solution is to distinguish between 'real' init and 'retry' init (with the IsRetry flag in my latest commit), and skip fetching offset if it is a retry.

id, line 438: Yes, sounds reasonable.

zmstone:
@id good to merge?
i'll add the callback brod_subscriber behaviour module in another pr to #17

id:
@zmstone I'll go through tomorrow morning once again.

PR #23, opened 2016-01-17
Preserve partition producer buffer

id:
In progress, not tested yet.


src/brod_producer_buffer.erl
```diff
@@ -34,26 +36,24 @@
...
         }).
 
--type send_fun() :: fun(([{binary(), binary()}]) -> {ok, corr_id()}).
--define(ERR_FUN, fun() -> erlang:error(bad_init) end).
-
 -record(buf,
-        { buffer_limit = 1        :: pos_integer()
-        , onwire_limit = 1        :: pos_integer()
-        , msgset_bytes = 1        :: pos_integer()
-        , send_fun     = ?ERR_FUN :: send_fun()
```
zmstone, line 51: i loved this send_fun...
could you please change it back ?
it makes tests way much easier.

src/brod_producer.erl
```diff
@@ -221,11 +232,23 @@ handle_info({msg, Pid, CorrId, #produce_response{} = R},
         "Offset: ~B\n"
         "Error code: ~p",
         [Topic, Partition, Offset, ErrorCode]),
-      {stop, {produce_error, {Topic, Partition, Offset, ErrorCode}}, State};
+      Error = {produce_response_error, Topic, Partition, Offset, ErrorCode},
+      {ok, NewState} = schedule_retry(State, Error),
```
zmstone: since we have a -1 retry times.
don't we want to be selective on errors to retry ?

src/brod_client.erl
```diff
@@ -109,23 +109,27 @@ stop(Client) ->
 %% @end
 -spec get_leader_connection(client(), topic(), partition()) -> {ok, pid()}.
```
zmstone: spec

src/brod_producer.erl
zmstone: and if leader is not available, or the it's not the leader we have connection to, we should re init the socket before schedule retry.
or rather, always re-init the socket (refresh metadata) for each retry?

id: Good point.

src/brod_producer.erl
```diff
@@ -247,31 +285,65 @@ terminate(_Reason, _State) ->
...
+  case brod_producer_buffer:maybe_send(Buffer, SockPid) of
+    {ok, Buffer} ->
+      {ok, State#state{buffer = Buffer}};
+    {error, Reason} ->
+      schedule_retry(State, Reason)
+  end;
+maybe_retry(_State) ->
+  false.
+
+is_retriable(ErrorCode) when ErrorCode =:= ?EC_CORRUPT_MESSAGE;
```
zmstone: why? error corrupted tcp packet when sending over the wire ?

src/brod_producer.erl
```diff
@@ -145,61 +160,68 @@ sync_produce_request(#brod_produce_reply{call_ref = CallRef} = Reply) ->
...
+    fun(SockPid, KafkaKvList) ->
+        Data = [{Topic, [{Partition, KafkaKvList}]}],
+        KafkaReq = #produce_request{ acks    = RequiredAcks
+                                   , timeout = AckTimeout
+                                   , data    = Data
+                                   },
+        case brod_sock:send(SockPid, KafkaReq) of
+          ok              -> ok;
+          {ok, CorrId}    -> {ok, CorrId};
+          {error, Reason} -> {error, Reason}
```
zmstone, line 186: Reason is always {sock_down, SockDownReason}.

 i don't see how it's possible to retry now neither in the future. why don't keep it simple ?

zmstone, line 186: never mind.
this is actually very retriable.

src/brod_producer.erl
```diff
@@ -247,31 +285,65 @@ terminate(_Reason, _State) ->
...
+    {error, Reason} ->
+      schedule_retry(State, Reason)
+  end;
+maybe_retry(_State) ->
+  false.
+
+is_retriable(ErrorCode) when ErrorCode =:= ?EC_CORRUPT_MESSAGE;
+                             ErrorCode =:= ?EC_UNKNOWN_TOPIC_OR_PARTITION;
+                             ErrorCode =:= ?EC_LEADER_NOT_AVAILABLE;
+                             ErrorCode =:= ?EC_NOT_LEADER_FOR_PARTITION;
```
zmstone: when 'not leader for partition' is returned, it means we need to refresh metadata. i.e. new socket.
or are we expecting the leader to move back before the retries run out of time?

zmstone: never mind, the socket is always reinitialised when retry.
i need sleep.

src/brod_producer.erl
```diff
@@ -214,16 +236,25 @@ handle_info({msg, Pid, CorrId, #produce_response{} = R},
...
+        true ->
+          {ok, NewState} = schedule_retry(State, Error),
+          {ok, NewBuffer} = brod_producer_buffer:nack(Buffer, CorrId),
+          {noreply, NewState#state{buffer = NewBuffer}};
+        false ->
+          {stop, Error, State}
+      end;
     false ->
-      NewBuffer = brod_producer_buffer:ack(Buffer, CorrId),
+      {ok, NewBuffer} = brod_producer_buffer:ack(Buffer, CorrId),
```
zmstone: i can't find a place where #state.retries is set back to 0.
should it be here ?

src/brod_producer_buffer.erl
```diff
@@ -34,26 +36,26 @@
...
 %% keep data in fun() to avoid huge log dumps in case of crash etc.
 -type data() :: fun(() -> kafka_kv()).
--type call() :: brod_call_ref().
 
 -record(req,
-        { call :: call()
-        , data :: data()
+        { call_ref :: brod_call_ref()
+        , data     :: data()
+        , bytes    :: non_neg_integer()
```
zmstone, line 47: would it make moe sense to have a failure_counter here?
the counter is bumped whenever SenFun returned error, or when it is 'nack'-ed.

then SendFun do an exit(LastError) (or return a non-retriable error Reason) when failure_count hits the limit.

comparing to a global counter in the producer looping state, this approaches is more robust as there is no need to reset the counter etc.

id, line 47: Good idea, go ahead.

src/brod_producer.erl
id: Yes.
https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-ErrorCodes

src/brod_producer.erl
id: oops

src/brod_producer_buffer.erl
id, line 47: No reason to fix 'retries' bug then.

zmstone, line 47: correct.

zmstone:
i think this is all i have for this pr. @id  go ahead merge it if you have no more comments.

test has been verified in my dev env. 
travis is really slow since yesterday.

id:
Awesome work. I'll go ahead and merge both this one and #17.

PR #31, opened 2016-02-23
add API to start per-topic producer dynamically

### LINKED ISSUE #30: Add API to start a topic producer on demand
Need this for REST Kafka service since we don't know upfront which topics users will be producing to.

id:
fixes #30


src/brod_client.erl
```diff
@@ -270,6 +277,22 @@ handle_info(Info, State) ->
                           [?MODULE, self(), State#state.client_id, Info]),
   {noreply, State}.
 
+handle_call({start_producer, TopicName, ProducerConfig}, _From, State) ->
+  case State#state.producers_sup of
+    undefined ->
```
zmstone: how about remove this clause: `maybe_start_producers_sup([])` and always have the supervisor started (no children if empty list)

id: :+1:

src/brod_client.erl
```diff
@@ -243,14 +250,13 @@ handle_info({'EXIT', Pid, Reason}, #state{ client_id     = ClientId
                                          } = State) ->
   error_logger:error_msg("client ~p producers supervisor down~nReason: ~p",
                          [ClientId, Pid, Reason]),
-  %% shutdown all producers?
-  {noreply, State#state{producers_sup = ?undef}};
+  {stop, {producers_sup_down, Reason}, State};
```
id, line 260: I see no reason to have client running when producers_sup or consumers_sup died. State is lost anyway.

src/brod_producers_sup.erl
```diff
@@ -55,6 +56,13 @@
 start_link(ClientPid, Producers) ->
   supervisor3:start_link(?MODULE, {?SUP, ClientPid, Producers}).
 
+%% @doc Dynamically start a per-topic supervisor
+-spec start_producer(pid(), pid(), topic(), producer_config()) ->
+                        {ok, pid()} | {error, any()}.
+start_producer(SupPid, ClientPid, TopicName, Config) ->
```
zmstone, line 62: we need a existence check on the topic name before starting the child.
or change it not to auto restart child on post-init failure.

id:
@zmstone more comments?

src/brod_client.erl
```diff
@@ -520,21 +576,21 @@ start_metadata_socket(_ClientId, [], Reason) ->
...
-maybe_start_producers_sup(Producers) ->
-  brod_producers_sup:start_link(self(), Producers).
-
-maybe_start_consumers_sup([]) ->
-  {ok, undefined};
-maybe_start_consumers_sup(Consumers) ->
-  brod_consumers_sup:start_link(self(), Consumers).
+-spec is_topic_exist(pid(), topic()) -> boolean().
+is_topic_exist(Socket, Topic) ->
+  Request = #kpro_MetadataRequest{topicName_L = []},
```
zmstone: why don't send just one topic name in the list ?
it can be a quite a lot of info for the broker to collect when trying to reply this request.

id: If auto.create.topics.enable set to true, this request will create that topic.

zmstone: then let it create, otherwise brod will not support topic auto-creation at all?

id: Which is not a bad thing me thinks. Auto-create topics is evil.

What if I add a flag in brod config to enable/disable topic existence verification (effectively enabling/disabling auto.create.topics support) ?

zmstone: evil or not should be the user's judgement call.
ok to add a flag.

zmstone:
good to merge

PR #35, opened 2016-03-07
add option to auto start producers

id:
This is mostly to simplify logic in REST proxy.
Might be useful for other users as well.


src/brod.erl
```diff
@@ -174,7 +174,19 @@ get_consumer(Client, Topic, Partition) ->
...
                        | {producer_not_found, topic(), partition()}.
 get_producer(Client, Topic, Partition) ->
-  brod_client:get_producer(Client, Topic, Partition).
+  case brod_client:get_producer(Client, Topic, Partition) of
+    {ok, Producer} ->
+      {ok, Producer};
+    {error, Reason} = Error
+      when Reason =:= {producer_not_found, Topic};
+           Reason =:= {producer_not_found, Topic, Partition} ->
+      %% try to start a producer for the given topic if
```
zmstone: when topic or partition not found, it is very likely a misconfiguration. why would we want to retry in this case ?

zmstone:
not quite sure about default producer config.

in my latest push consumer-group branch, i added a check to return ok directly when the producer is found (cost is a ets lookup).
meaning, the callers can always try to start client before producing messages.

id:
Scenario: my application does not know in advance which topics it will produce data to. It gets requests and calls brod to produce data to Kafka.
Without this feature it will have to:
- call brod:get_producer to verify if a producer for that partition is already launched
- if yes, produce
- if not, call brod:start_producer, call brod:get_producer again and produce if it was not a error

Situation gets more complicated if the application will need to use custom partitioner. It would have to check whether a producer for a given topic is running (which means a topic supervisor, we don't have api for this yet), and then call a regular brod:produce.

With this feature an application can just enable auto_start_producers option for the client and keep using brod:produce.

src/brod.erl
```diff
@@ -174,7 +174,19 @@ get_consumer(Client, Topic, Partition) ->
                        | {producer_not_found, topic()}
                        | {producer_not_found, topic(), partition()}.
 get_producer(Client, Topic, Partition) ->
-  brod_client:get_producer(Client, Topic, Partition).
+  case brod_client:get_producer(Client, Topic, Partition) of
+    {ok, Producer} ->
+      {ok, Producer};
+    {error, Reason} = Error
+      when Reason =:= {producer_not_found, Topic};
+           Reason =:= {producer_not_found, Topic, Partition} ->
```
zmstone: this error means topic is found, but there is no such partition, should not  try to start a producer.

id: Right.

id:
ping @zmstone

PR #42, opened 2016-04-07
Fix typos in README.md

sdball:
There are a couple of typos at the top of the current 2.0 README that I've fixed in this PR.


zmstone:
thanks

PR #43, opened 2016-04-10
New connection manager

juise:
A little bit refactored network layer

### DIFF
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -146,10 +146,10 @@ start_producer(Client, TopicName, ProducerConfig) ->
 %% @doc Dynamically start a topic consumer.
 %% @see brod_client:start_consumer/3. for more details.
 %% @end
--spec start_consumer(client(), topic(), producer_config()) ->
+-spec start_consumer(client(), topic(), consumer_config()) ->
                         ok | {error, any()}.
-start_consumer(Client, TopicName, ProducerConfig) ->
-  brod_client:start_consumer(Client, TopicName, ProducerConfig).
+start_consumer(Client, TopicName, ConsumerConfig) ->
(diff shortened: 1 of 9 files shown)


zmstone:
could you elaborate the motivation behind refactoring brod_sock ?

id:
I double the question from @zmstone.
The reasoning behind using custom process for brod_sock was to 1) have better control on how we process messages and 2) avoid gen_server overhead.

id:
Also please PR to 2.0-dev, consumer-group branch has been already merged.

juise:
> could you elaborate the motivation behind refactoring brod_sock ?

Sure! I'm started working on https://github.com/klarna/brod/issues/40, but understand that I know not so much about library internals. The current impl of brod_sock is OTP-like but complicated for understand, so I decided to make it more simple!

> The reasoning behind using custom process for brod_sock was to 1) have better control on how we process messages and 2) avoid gen_server overhead.
1. As I know, you can get the same control level using gen_server.
2. Could you please explain about gen_server overhead? I current impl of brod_sock use the same mechanisms for send and receive messages.

> Also please PR to 2.0-dev, consumer-group branch has been already merged.

If I close current PR and make new to 2.0-dev, is it will be okay?

id:
> so I decided to make it more simple!

It implements only those pieces of a regular erlang process that are required by the library + supports the most important system messages. It's way simpler than gen_server.erl :)

1,2. Try to follow handle_msg in https://github.com/erlang/otp/blob/maint/lib/stdlib/src/gen_server.erl#L660, there is a lot of logic happening before and after callbacks are called. We don't need all of it.

> If I close current PR and make new to 2.0-dev, is it will be okay?

Of course!

PR #44, opened 2016-04-11
New connection manager

juise:
Ref: https://github.com/klarna/brod/pull/43

### DIFF
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -146,10 +146,10 @@ start_producer(Client, TopicName, ProducerConfig) ->
 %% @doc Dynamically start a topic consumer.
 %% @see brod_client:start_consumer/3. for more details.
 %% @end
--spec start_consumer(client(), topic(), producer_config()) ->
+-spec start_consumer(client(), topic(), consumer_config()) ->
                         ok | {error, any()}.
-start_consumer(Client, TopicName, ProducerConfig) ->
-  brod_client:start_consumer(Client, TopicName, ProducerConfig).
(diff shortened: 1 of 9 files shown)


juise:
> 1,2. Try to follow handle_msg in https://github.com/erlang/otp/blob/maint/lib/stdlib/src/gen_server.erl#L660, there is a lot of logic happening before and after callbacks are called. We don't need all of it.

I looked at gen_server and all the code before and after callback is clean and briliant, and do the same what old brod_sock doing: loop, dispatch, and handle message. But right now gen_server does it all for you!

id:
I would still prefer to keep brod_sock as is.

juise:
Hm, why, the perfomance is not degrade, but code is more clean and strict!

zmstone:
while we are continuing the discussion about brod_sock rewrites.
I would love to merge the typo fixes and the renaming of send/send_sync to request_async/request_sync NOW.
@juise could you split that part to another pr?

also, travis is not happy because of some edoc failure.

id:
Current code is simple and clean, it is doing exactly what it has to do and nothing more, it is a verified implementation. Switching to gen_server is a step back in my opinion.

Take a look at http://ninenines.eu/talks/beyond-otp/beyond-otp.html if you are not convinced.

juise:
> I would love to merge the typo fixes and the renaming of send/send_sync to request_async/request_sync NOW.
> @juise could you split that part to another pr?

@zmstone Sure! I do it tomorrow in separate PR!

PR #45, opened 2016-04-12
fix typos, more declarative brod_sock api

juise:


src/brod_group_subscriber.erl
```diff
@@ -22,7 +22,7 @@
 %%%
 %%% An overview of what it does behind the scene:
 %%%  1. Start a consumer group controller,
-%%%     @see brod_group_controller:start_link/4. to manage the consumer group
+%%%     @see brod_group_controller:start_link/4 to manage the consumer group
```
id: The dot here was added intentionally to satisfy Edoc, without dot it fails.

juise:
Ouch!

Do you mind, if i rephrase? - For manage the consumer group state, start a consumer group controller, @see brod_group_controller:start_link/4.

id:
May be even "Start a consumer group controller to manage the consumer group states: @see ..." ?

id:
Very nice, thanks!

PR #51, opened 2016-04-17
consumer lazy payload socket usage

### LINKED ISSUE #41: be lazy on establishing payload connections
in current (2.0-dev) implementation: if an idle payload connection is killed by kafka (or some load balancer etc.), brod_consumer will try to setup a new one after a (configurable) timeout.
this is not necessary when there is currently no subscriber listening.

it should allow subscribers to subscribe when there is no alive brod_sock (instead of return {error, no_connection}), and try to connect to the partition leader _after_ the subscriber checks in.

zmstone:
consumer should not alway try to re-connect to partition leader if there is no subscriber.
fixes #41


id:
💯

PR #54, opened 2016-04-21
Add supervisor3 and kafka_protocol apps to brod.app.src

ates:


id:
Any special reason for this?
It will not hurt of course to have them here, but both supervisor3 and kafka_protocol are libraries, they not need to be started.

ates:
rebar3 don't include these libraries during release creation, as result application can't start:

```
Erlang/OTP 18 [erts-7.3] [source] [64-bit] [smp:8:8] [async-threads:10] [kernel-poll:true]


=INFO REPORT==== 21-Apr-2016::22:47:07 ===
    application: brod
    exited: {bad_return,
                {{brod,start,[normal,[]]},
                 {'EXIT',
                     {undef,
                         [{supervisor3,start_link,
                              [{local,brod_sup},brod_sup,clients_sup],
                              []},
                          {application_master,start_it_old,4,
                              [{file,"application_master.erl"},
                               {line,273}]}]}}}}
    type: permanent
{"Kernel pid terminated",application_controller,"{application_start_failure,brod,{bad_return,{{brod,start,[normal,[]]},{'EXIT',{undef,[{supervisor3,start_link,[{local,brod_sup},brod_sup,clients_sup],[]},{application_master,start_it_old,4,[{file,\"application_master.erl\"},{line,273}]}]}}}}}"}
```

id:
Oh. Ok, I see. Thanks!

PR #56, opened 2016-04-23
check error code when validating topic existence

### LINKED ISSUE #55: Different behavior when producing to a random partition or to a specified one
When producing to a random partition with autostart producer an error is returned in case topic doesn't exist, which is ok.

When producing to a specified topic AND specified partition with producer autostart the process is crashing instead of reporting that topic or/and partition is not found.

zmstone:
fixes #55


zmstone:
@id @kirillzhiganov call for review

id:
👍

PR #66, opened 2016-05-17
fix buffer issue when required acks=0

### LINKED ISSUE #65: required_acks=0 only processes first message
Hi,

I've just started exploring the driver. I notice that when required_acks = 0 (which if I'm not mistaken basically tells kafka to not send any acknowledgments), only the first message gets processed by the consumer. All succeeding messages are not consumed. 

Upon further investigation, it seems as if the driver is stuck on expecting a response or something? I'm not 100% sure as I have to delve into abit of code. But just incase you guys have any idea? 

I'm basically just calling

brod:produce(brod1, ?SESSION_TOPIC, ?RANDOM_PARTITION,  <<>>, J).

sync_produce will result in permanent waiting of response.

Please let me know.

Thanks!
Byron

zmstone:
fixes #65

### DIFF
--- a/src/brod.app.src
+++ b/src/brod.app.src
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Apache Kafka Erlang client library"},
-  {vsn,"2.0.0"},
+  {vsn,"2.0.1"},
   {registered,[]},
   {applications,[kernel,stdlib,supervisor3,kafka_protocol]},
   {env,[]},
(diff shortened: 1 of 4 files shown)


zmstone:
@id after an attempt to make brod:produce api return ok instaed of {ok, CallRef} i decided to give up for 2 reasons:
1. we have retries even when required_acks = 0 when socket is down, in such case, caller may continue buffering, and sync later on (e.g. when socket is back).
2. it's to much special handing for the 'buffered' and 'acked' message handing in brod_producer internal.
perhaps not worth it to trade this small user-convenience for clean code.

id:
Agree.
When required_acks == 0, clients should discard return result from this call anyway.

PR #74, opened 2016-05-20
Fix specs: clients ids should have type brod_client_id() (i.e. atom()), not …

SimonInman:
…client_id() (which is string(), and only exists in the kafka_protocol library in any case)

This was causing a dialyzer error when I attempted to use the client. I believe the problem arose due to confusion between client_id() and brod_client_id(). I think all references within the brod code should be to brod_client_id(), and I think this PR makes all the needed changes.


zmstone:
Thanks for reporting this @SimonInman 
We did not run any dialyzer check on brod.
found some more issues when I try to run the checks, will fix in this pr based on your commit.

zmstone:
merged. will send another pr for more fixes. thanks! @SimonInman

PR #75, opened 2016-05-21
Fix dialyzer checks

zmstone:


id:
Guess we need to wait for another test run when PRs to supervisor3 and kafka_protocol are merged.
Otherwise good to go.

PR #80, opened 2016-05-30
brod_sock exits with 'normal' reason on tcp_closed

id:
this is to avoid crash report from proc_lib


id:
@zmstone ping

zmstone:
LGTM, @byronpc any comment ?

```
=INFO REPORT==== 31-May-2016::19:07:54 ===
client aaa metadata socket down localhost:9092
Reason:{shutdown,tcp_closed}
```

PR #81, opened 2016-06-01
remove brod_client:start_link/2

zmstone:
All public APIs (exposed from brod.erl) call brod_client:start_link/3.
There is no need to keep brod_client:start_link/2 any more.
Also added few TODOs to remind us to clean up more when the deprecated APIs brod:start_link_client/_ are removed


id:
👍

PR #83, opened 2016-06-05
Better handling of leader errors

id:


src/brod_utils.erl
```diff
@@ -142,17 +142,23 @@ ensure_integer_offset_time(T) when is_integer(T) -> T.
...
-    lists:keyfind(Partition, #kpro_PartitionMetadata.partition, Partitions),
-  Id >= 0 orelse erlang:throw({no_leader, {Topic, Partition}}),
+  Id = case lists:keyfind(Partition,
+                          #kpro_PartitionMetadata.partition, Partitions) of
+         #kpro_PartitionMetadata{leader = Leader} when Leader >= 0 ->
+           Leader;
+         #kpro_PartitionMetadata{} ->
+           erlang:throw(?EC_LEADER_NOT_AVAILABLE);
+         _ ->
+           erlang:throw(?EC_UNKNOWN_TOPIC_OR_PARTITION)
```
zmstone, line 162: here we know for sure it's an unknown partition, stating 'topic or partition' will be misleading.
perhaps invent an error unknown_partition ?

id, line 162: Then users will have more different kinds of errors to care about. I'd rather keep it simple.

PR #89, opened 2016-07-22
support produce kv-list batch as message set

zmstone:
For messages that are already batched on caller side, there is no need to break down and re-assemble.
Also easier to achieve better compression ratio if compression is enable.

### DIFF
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -99,11 +99,14 @@
              , endpoint/0
              , group_config/0
              , group_id/0
+             , key/0
              , kpro_MetadataResponse/0
+             , kv_list/0
              , offset/0
              , offset_time/0
              , producer_config/0
              , topic/0
+             , value/0
              ]).
 
 -include("brod_int.hrl").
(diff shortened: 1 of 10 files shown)


src/brod.erl
```diff
@@ -248,22 +251,30 @@ get_producer(Client, Topic, Partition) ->
...
-%% @doc Produce one message. The pid should be a producer pid.
--spec produce(pid(), binary(), binary()) ->
+%% @doc Produce one message if Value is binary or iolist,
+%% or a message set if Value is a (nested) kv-list, in this case Key
+%% is discarded (only the keys in kv-list are sent to kafka).
+%% The pid should be a partition producer pid, NOT client pid.
+%% @end
+-spec produce(pid(), key(), value()) ->
         {ok, brod_call_ref()} | {error, any()}.
 produce(ProducerPid, Key, Value) ->
```
id, line 266: Perhaps exclude this function from brod: api to not confuse with produce call for brod client?
If a user knows what they are doing, they can just use brod_producer directly.

id, line 266: @zmstone let me know if you agree with this and I'll fix it

zmstone: that should be a 3.0 change

id: All right, approved.

Issue #76, opened 2016-05-22
Feature Request: TLS Encryption


varnerac:
Kafka 0.9 supports TLS encryption.

http://www.confluent.io/blog/apache-kafka-security-authorization-authentication-encryption

id:
Fixed by #93

PR #93, opened 2016-08-19
support tls connection

id:


id:
Fixes #76

src/brod_sock.erl
```diff
@@ -139,28 +150,44 @@ debug(Pid, File) when is_list(File) ->
...
+                [] ->
+                  State0#state{mod = gen_tcp, sock = Sock};
+                SslOpts ->
+                  error_logger:info_msg("Trying to establish ssl connection with options: ~p\n", [SslOpts]),
+                  case ssl:connect(Sock, SslOpts, Timeout) of
+                    {ok, SslSock} ->
+                      State0#state{mod = ssl, sock = SslSock};
+                    {error, Reason} ->
+                      exit({ssl_negotiation_failure, Reason,
+                            erlang:get_stacktrace()})
```
zmstone: why stacktrace here ?

id: to get more detailed info in crash log

zmstone:
looks good

PR #94, opened 2016-09-02
rebar2 compatibility

brigadier:
`rebar.config.script` file to make it possible to compile the app with rebar2.


id:
LGTM. @zmstone ?

rebar.config.script
```diff
@@ -0,0 +1,12 @@
+IsRebar3 = erlang:function_exported(rebar3, main, 1),
+Rebar2Deps = [
```
zmstone: could you take the version tags from rebar.config instead of repeating here ?

brigadier:
> could you take the version tags from rebar.config instead of repeating here ?

done

zmstone:
good stuff!
thanks!

PR #100, opened 2016-09-16
shrink max_bytes

zmstone:


id (commented):
What is so wrong in having max_bytes equal to the size of the largest message all the time that we should try to avoid it by doing computations on every receive?

If you really think it does worth it, perhaps it makes sense to add a clause for maybe_shrink_max_bytes which does nothing if max_bytes == max_bytes_orig.

id:
Congrats with a round PR number! :) 💯

zmstone:
The idea is to 'predict' the next batch size when shrinking
In some cases, MaxMessageBytes in history might still be too big for regular messages.
e.g. few 10M messages appear every once in a while in a regularly 1K message stream.
The subscriber may starve while waiting for the consumer to fetch such a big batch.
(unless the prefetch count is large enough)

Regarding

> add a clause for maybe_shrink_max_bytes which does nothing if max_bytes == max_bytes_orig

This would imply that the avg_bytes is only calculated in the message set right after a successful expansion, which would over estimate the average value.

zmstone:
This is also to allow a really small max_bytes in config, and slowly adjust to a 'smart' number to fetch a predictable number of messages in each request.
But we need to also respect a large max_bytes in config (not to shrink to a lower value)

id (approved):

PR #101, opened 2016-09-26
Fix kafka_protocol version in rebar.config

ruanjonker:


zmstone:
already fixed in 2.2-rc1

PR #102, opened 2016-09-28
brod_sock: if requiredAcks = 0 don't store the request

eriklee:
Hi,
First off, thanks for all your work on the library. We're investigating using kafka for logging at my job and while trying to set that up I've noticed a few issues. I realize that since we're looking for high-throughput logging our goals might be slightly different from yours, but hopefully most of our changes will be useful for everyone using the library.

I noticed that the brod_sock state will grow indefinitely if requiredAcks is set to 0 because we keep them while waiting for a response. I presume we still need to increase the corr_id so I've added a new function to do that without storing the request.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT = brod
 PROJECT_DESCRIPTION = Kafka client library in Erlang
-PROJECT_VERSION = 2.1.9
+PROJECT_VERSION = 2.1.10
 
 DEPS = supervisor3 kafka_protocol
 
--- a/src/brod.app.src
+++ b/src/brod.app.src
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Apache Kafka Erlang client library"},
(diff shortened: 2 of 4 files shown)


eriklee:
Hmm, the build failure doesn't look like something I would have caused:
src/brod_group_subscriber.erl:40: behaviour brod_group_member undefined

Should this be merged into the 2.2-dev branch instead?

zmstone:
@eriklee Thanks for the PR.

> src/brod_group_subscriber.erl:40: behaviour brod_group_member undefined

I believe the above error is caused by the `-S` erlc flag, why do you need this ?

zmstone:
please bump version in Makefile and app.src to 2.1.10

eriklee:
Sorry, the compiler flag snuck in while I was trying to split changes out into multiple branches.

zmstone:
please rebase on latest master and squash the last two commits.

zmstone (approved):

id:
Good stuff, thanks!

PR #105, opened 2016-09-29
brod_producer_buffer: use a queue for pending and buffer

eriklee:
This switches the brod_producer_buffer pending and buffer state elements to be queues instead of lists. This seems to have improved performance with heavy load a fair bit.

Unfortunately it does make the code a fair bit uglier, but I tried to avoid it as much as I could with the EMPTY_QUEUE macro.

### DIFF
--- a/src/brod_producer_buffer.erl
+++ b/src/brod_producer_buffer.erl
@@ -54,17 +54,19 @@
                         {error, any()}).
 -define(ERR_FUN, fun() -> erlang:error(bad_init) end).
 
+-define(EMPTY_QUEUE, {[],[]}).
+
 -record(buf,
-        { buffer_limit   = 1        :: pos_integer()
-        , onwire_limit   = 1        :: pos_integer()
-        , max_batch_size = 1        :: pos_integer()
-        , max_retries    = 0        :: integer()
-        , send_fun       = ?ERR_FUN :: send_fun()
-        , buffer_count   = 0        :: non_neg_integer()
-        , onwire_count   = 0        :: non_neg_integer()
-        , pending        = []       :: [#req{}]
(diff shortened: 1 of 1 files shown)


zmstone:
I think you have missed some lines. e.g. line 278.
also, run ct with: `make ct-brod_producer_buffer`

zmstone:
@id  do you think it's time to stop supporting pre-17 OTP?
otherwise we'll have to add some compilation switches to support queue() or queue:queue() type specs (both for erlang.mk and rebar).

id:
@zmstone yep, I think we should drop it.

eriklee:
Ah, so I have. Apparently my brokers never fail :)

Now passes the tests. I'm having a hard time getting `make test-env t` to work (lots of problems with timeouts and sock down). ~~Is this expected?~~ I see they are from the official one. I'm still getting non-0 error codes though...

zmstone:
Good work!

Regarding tests.
First, there are intended failures in the SUITEs (in case you are referring to those kind of timeouts and sock down).

We have seen unstable tests in travis, sometimes I have to rerun the tests, but mostly due test-env failure.

In local setup, you may run `make test-env` first, then repeat `make t`.
I've noticed consumer group failures in the initial run, but usually get back to normal in the second run.

If you notice any test case fail repeatedly, please send an issue to us.

zmstone (approved):

id (approved):

zmstone:
@eriklee Thanks a lot, again! :)
Please report an issue for your non-0 error codes in `make t` if it's not a problem in your local setup.

eriklee:
Found the problem, there was a bad version of kafka_protocol in my deps. Everything seems to be working now. :)

PR #110, opened 2016-10-05
Simplify subscribers

zmstone:


src/brod_group_subscriber.erl
```diff
@@ -389,55 +357,51 @@ terminate(_Reason, #state{}) ->
...
-        NewConsumer = Consumer#consumer{acked_offset = Offset},
-        NewConsumers = lists:keyreplace({Topic, Partition},
-                                        #consumer.topic_partition,
-                                        Consumers, NewConsumer),
-      State0#state{consumers = NewConsumers};
+  case lists:keyfind({Topic, Partition},
+                     #consumer.topic_partition, Consumers) of
+    #consumer{consumer_pid = ConsumerPid} = Consumer ->
+      ok = brod:consume_ack(ConsumerPid, Offset),
+      ok = brod_group_coordinator:ack(Coordinator, GenerationId,
```
id, line 399: would it make sense to batch acks per partition and ack the last one?

zmstone: NO, for 2 reasons:
1. We do not know the cost of each `CbModule:handle_message/2`, in case it's slow, acking only the last one will delay the pre-fetches.
2. `brod:consume_ack/2` and `brod_group_coordinator:ack/5` are quite cheap.

id: Ok, fair enough.

id (approved):

id (approved):

PR #111, opened 2016-10-07
Fix SSL on Consumer Group

eriosv:
Hi,

I've done some testing connecting a Brod service with a Kafka with an SSL listener (no plaintext listener). Producing messages works perfectly but the consumer group is failing.

The problem is that the socket opened by the brod_group _coordinator is not using the SSL configuration defined in the client.

### DIFF
--- a/src/brod_client.erl
+++ b/src/brod_client.erl
@@ -235,7 +235,7 @@ get_partitions_count(Client, Topic) when is_pid(Client) ->
 
 %% @doc Get the endpoint of the group coordinator broker.
 -spec get_group_coordinator(client(), group_id()) ->
-        {ok, endpoint()} | {error, any()}.
+        {ok, {endpoint(), client_config()}} | {error, any()}.
 get_group_coordinator(Client, GroupId) ->
   safe_gen_call(Client, {get_group_coordinator, GroupId}, infinity).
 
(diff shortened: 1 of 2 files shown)


id (approved):
Good stuff, thanks!

zmstone:
Sorry, I thought I've sent the comments, but turns out I didn't (not quite familiar with github's new review process).
My comment was: `brod_client:get_group_coordinator` can return client config, so there is no need to make another gen_server call to get the config.

zmstone:
Also, I'd love to have the lines limited within 80-char width

src/brod_client.erl
```diff
@@ -387,6 +387,8 @@ handle_call({get_metadata, Topic}, _From, State) ->
 handle_call({get_connection, Host, Port}, _From, State) ->
   {NewState, Result} = do_get_connection(State, Host, Port),
   {reply, Result, NewState};
+handle_call(get_config, _From, State) ->
```
zmstone: this clause is no longer required.

eriosv: sure... missed that one

zmstone:
last one: please squash the commits

zmstone:
Thanks!

eriosv:
Thank you!

PR #112, opened 2016-10-08
Simplify subscribers for 2.1

zmstone:


id (approved):
I'd like to see more examples in README, but we can do this later. Perhaps even put this in the project wiki.

id (approved):

PR #129, opened 2016-10-22
rename second level per-topic supervisor MACROs

zmstone:
for producers
SUP -> PRODUCERS_SUP
SUP2 -> PRODUCER_SUP
for consumers
SUP -> CONSUMERS_SUP
SUP2 -> CONSUMER_SUP
keeping the actual name unchanged (e.g. brod_topics_supervisor2)
to be backword compatible. otherwise a full supervision tree
restart is required when load recompiled beam.

### DIFF
--- a/src/brod_consumers_sup.erl
+++ b/src/brod_consumers_sup.erl
@@ -34,11 +34,11 @@
 
 -include("brod_int.hrl").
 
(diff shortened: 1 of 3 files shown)


id (approved):

zmstone:
no, this is not good naming either. closing it.

PR #139, opened 2016-11-25
fix off-by-one error check

beano:
Consuming the first message isn't possible due to an off-by-one error in an
input check.


zmstone:
duplicate to #140

PR #140, opened 2016-11-25
fix off-by-one error check

beano:
Consuming the first message isn't possible due to an off-by-one error in an
input check.


zmstone:
could you elaborate on why you need this fix?
i'm not quite following because 'acked' offsets should be 'real' offsets, which is zero based.

zmstone:
ah, sorry, missed the pr description.
i think in this case, you need to provide [{begin_offset, earliest}] as consumer config in start_consumer call

zmstone:
not an issue.
the right way to consume from earliest offset is not by faking a committed offset, (e.g. `-2` for earliest), but to provide a default `beging_offset` in consumer config.

PR #151, opened 2016-12-25
Use erlang:throw instead of erlang:error in do_commit_offsets_

MikhailKalashnikov:
Detaiils https://github.com/klarna/brod/issues/150

### DIFF
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -682,34 +682,34 @@ do_commit_offsets_(#state{ groupId                  = GroupId
       },
   Rsp = send_sync(SockPid, Req),
   #kpro_OffsetCommitResponse{oCRspTopic_L = Topics} = Rsp,
-  ErrorSet = lists:foldl(
-    fun(#kpro_OCRspTopic{topicName = Topic, oCRspPartition_L = Partitions},
-        Acc1) ->
-      lists:foldl(
-        fun(#kpro_OCRspPartition{partition = Partition, errorCode = EC},
-            Acc2) ->
-          case kpro_ErrorCode:is_error(EC) of
-            true -> 
-              log(State, error,
-                "failed to commit offset for topic=~s, partition=~p\n"
-                "~p:~s", [Topic, Partition, EC, kpro_ErrorCode:desc(EC)]),
-              gb_sets:add_element(EC, Acc2);
-            false -> Acc2
-          end
-        end, Acc1, Partitions)
-    end, gb_sets:new(), Topics),
-  %% if all error codes are the same, raise throw, otherwise error
-  case gb_sets:size(ErrorSet) of
-    0 -> ok;
-    1 -> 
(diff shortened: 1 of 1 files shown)


src/brod_group_coordinator.erl
```diff
@@ -682,19 +682,32 @@ do_commit_offsets_(#state{ groupId                  = GroupId
       },
   Rsp = send_sync(SockPid, Req),
   #kpro_OffsetCommitResponse{oCRspTopic_L = Topics} = Rsp,
-  lists:foreach(
-    fun(#kpro_OCRspTopic{topicName = Topic, oCRspPartition_L = Partitions}) ->
-      lists:foreach(
-        fun(#kpro_OCRspPartition{partition = Partition, errorCode = EC}) ->
-          kpro_ErrorCode:is_error(EC) andalso
-            begin
+  ErrorSet = lists:foldl(
```
zmstone: please move this whole block to a new function, something like this:

```erlang
assert_commit_response(Topics) ->
  ErrorSet = collect_commit_respons_error_codes(Topics)
  case gb_sets:to_list(ErrorSet) of
    [] -> ok;
    [EC] -> erlang:throw(EC);
    _ -> erlang:error({commit_offset_failed, Topics})
  end.
```

src/brod_group_coordinator.erl
```diff
@@ -682,19 +682,32 @@ do_commit_offsets_(#state{ groupId                  = GroupId
...
+          case kpro_ErrorCode:is_error(EC) of
+            true -> 
               log(State, error,
-                  "failed to commit offset for topic=~s, partition=~p\n"
-                  "~p:~s", [Topic, Partition, EC, kpro_ErrorCode:desc(EC)]),
-              erlang:error(EC)
-            end
-        end, Partitions)
-    end, Topics),
+                "failed to commit offset for topic=~s, partition=~p\n"
```
zmstone: no need to keep this log when collection error codes.
the 'throw' error code is logged when trying to stabilize.
the 'error' error codes will be a part of crash log.

src/brod_group_coordinator.erl
zmstone: oh, please use the ESCALATE_EC macro instead of erlang:throw.
-- to make code search easier.

src/brod_group_coordinator.erl
MikhailKalashnikov: But if we have different error codes, this information will be useful. What do you think?

zmstone: to keep it simple: send all error codes to crash log (as in my example code).
I'm not even sure if kafka would in practical return different error codes.

src/brod_group_coordinator.erl
MikhailKalashnikov: I've changed code as you suggested

zmstone:
very nice!
@MikhailKalashnikov thank you!

PR #152, opened 2016-12-26
Coordinator handle commit offset error

### LINKED ISSUE #150: Handle error in brod_group_coordinator commit_offsets
In brod_group_coordinator in handle_info(?LO_CMD_COMMIT_OFFSETS and in handle_call(commit_offsets you try to catch throw and stabilize.
But in do_commit_offsets_ you raise exception with erlang:error(EC), but not throw. 
As a result this error is not catched and server crashes without attempt to stabilize.
Is it correct/expected behavior or we should catch this kafka error?

zmstone:
fixes #150

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT = brod
 PROJECT_DESCRIPTION = Kafka client library in Erlang
-PROJECT_VERSION = 2.2.9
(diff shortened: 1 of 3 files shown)


id:
2.2.10?

id (approved):

Issue #155, opened 2017-01-30
produce rate is decreasing a lot when leader is changed/restarted


silviucpp:
Hello,

We observed in production when we restarted the leader some strange behavior where the produced message rate decreased a lot and all caller processes for produce method were blocked in that call.

Also I did a small script to send messages and after we restart the leader or try to move partition on another machine we can see the same behavior. It's taking over 30 seconds to get everything back to normal.

https://dl.dropboxusercontent.com/u/30548263/test_kafka.erl

     test_kafka:init().
     test_kafka:run_infinite(8, 10000, 0).

This will send in loop using 8 erlang processes messages to kafka on topic 0. If you look in observer (will be opened as well) after you restart the topic the messages queue for the brod processes are staying freezed and are cleaned after 30 -40 seconds. Also if you look in kafka the number of messages produced is decreasing a lot.

Any technical reason for this ? or any missed configure on our side ?

Silviu

zmstone:
Hi @silviucpp 

A part of this is a design intention (right or wrong is open for discussion).
Up on a failure, brod_producer will re-queue the failed messages and start over.
The retry is implemented in a 'slow' fashion: It sends sync-requests for one message at a time.
For two reason
1. To start slow (Looking at it retrospectively now, It's probably not necessary, will send a PR to fix it)
2. For message strict ordering

BTW. Leader re-discovery might have contributed a bit to your 30-40 seconds?

As a work around before we can fix it with a better batching, you can:
1. Batch on the caller side: `call brod:prodcue(Client, Topic, Patition, _Key = <<>>, _Value = [{Key, Value}])`:  so brod_producer can actually retry a batch (instead of one message) at a time.
2. Fail fast if the caller has the ability to retry (or if it is ok to have few messages getting lost): brod_producer by default has 3 reties, if `max_retries` given in producer config is 0, brod_producer will crash immediately, then supervisor should restart it (after a configurable sleep).

silviucpp:
Hello,

Leader discovery I don't think has any contribution as time I can see it's discovered almost instantly. 

Do you have any idea when an improvement in this are is planned ? In the meantime I will try your second suggestion

Silviu

zmstone:
I will send the pr in a day or two.

silviucpp:
I just tested with {max_retries, 0} as well now but I still see that poor behavior (low rate, caller blocked for a ling time). 
[...]

PR #156, opened 2017-01-30
Batch on retry

zmstone:
'retry slow' and 'retry one at a time' seem to be over engineered.
As reported in #155, (especially when the message sizes are small and buffer size is large), it may take quite long time to recover.

As for strict ordering, as long as onwire_limit is larger than 1, the risk of breaking it is always there.
I might have missed something though, @id please check me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT = brod
 PROJECT_DESCRIPTION = Kafka client library in Erlang
-PROJECT_VERSION = 2.2.10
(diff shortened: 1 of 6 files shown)


id:
You are right, I see the same.

id:
Something with tests?

zmstone:
Something wrong with my local testbed.
Had to retry a lot in travis.

id (approved):

PR #159, opened 2017-02-09
Do not send new fetch request before last response is received

### LINKED ISSUE #157: maybe_send_fetch_request function may send multi fetch request with the same begin_offset if prefetch_count is more than 1
when I set prefetch_count with 200, this may happen in this order:
1. when ack offset 1, function 'maybe_send_fetch_request' will send a fetch request
2. if ack offset 2 happened before receive the fetch response, function 'maybe_send_fetch_request' will send a fetch request with the same begin_offset
3. this may cause fetch request for the same begin_offset be sent for 200 times.

zmstone:
fixes #157

### DIFF
--- a/src/brod_consumer.erl
+++ b/src/brod_consumer.erl
@@ -65,7 +65,7 @@
                , max_bytes_orig      :: bytes()
                , sleep_timeout       :: integer()
                , prefetch_count      :: integer()
-               , last_corr_id        :: corr_id()
+               , last_corr_id        :: ?undef | corr_id()
                , subscriber          :: ?undef | pid()
                , subscriber_mref     :: ?undef | reference()
                , pending_acks = []   :: [offset()]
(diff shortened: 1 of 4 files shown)


src/brod_consumer.erl
```diff
@@ -301,8 +301,9 @@ handle_fetch_response(_Response, CorrId1,
                             } = State) when CorrId1 =/= CorrId2 ->
   {noreply, State};
 handle_fetch_response(#kpro_FetchResponse{ fetchResponseTopic_L = [TopicData]
-                                         }, CorrId, State) ->
-  CorrId = State#state.last_corr_id, %% assert
+                                         }, CorrId, State0) ->
+  CorrId = State0#state.last_corr_id, %% assert
+  State = State0#state{last_corr_id = ?undef},
```
id, line 307: shouldn't we also do this for `subscriber = ?undef` clause?

id (approved):

PR #164, opened 2017-02-15
change pending_acks in brod_consumer.erl from list to pending_acks record

theorix:


src/brod_consumer.erl
```diff
@@ -442,17 +476,50 @@ handle_reset_offset(#state{offset_reset_policy = Policy} = State, _Error) ->
...
+                        , count = Count} = PendingAcks, Offset) ->
+  {{value,{Begin,End}}, Queue1} = queue:out(Queue),
+  if
+    Offset >= End ->
+      NewCount = Count - (End - Begin + 1),
+      handle_ack(PendingAcks#pending_acks{ offsets_queue = Queue1
+                                         , count = NewCount
+                                         }, Offset);
+    Offset >= Begin ->
+      NewCount = Count - (Offset - Begin + 1),
```
zmstone: Yeah, so there is a need for queue:in_r when handling ack.
Perhaps it makes more sense to keep a queue head out of queue:queue() too?

zmstone:
maybe create a new module, call it `brod_consumer_offset_queue`?

src/brod_consumer.erl
```diff
@@ -352,13 +358,41 @@ handle_message_set(#kafka_message_set{messages = Messages} = MsgSet,
...
                       },
   State1 = maybe_shrink_max_bytes(State, MsgSet#kafka_message_set.messages),
   NewState = maybe_send_fetch_request(State1),
   {noreply, NewState}.
 
+handle_add_offset(#pending_acks{ queue_rear = {Begin, End}
+                               , count = Count
+                               , offsets_queue = Queue
+                               } = PendingAcks, [Offset|Offsets]) ->
+  if
```
zmstone: Sorry, but we don't like `if`, please change it to `case`

src/brod_consumer.erl
theorix: good, I will add it

src/brod_consumer.erl
theorix: ok

zmstone:
on a second thought, actually, this might be over engineered, 
let's keep it simple and use queue:queue() only.

src/brod_consumer.erl
```diff
@@ -352,13 +358,41 @@ handle_message_set(#kafka_message_set{messages = Messages} = MsgSet,
...
+    Offset == End + 1 ->
+      handle_add_offset(PendingAcks#pending_acks{ queue_rear = {Begin, Offset}
+                                                , count = Count + 1
+                                                }, Offsets);
+    Offset > End + 1 ->
+      NewQueue = queue:in({Begin, End}, Queue),
+      handle_add_offset(PendingAcks#pending_acks{ queue_rear = {Offset, Offset}
+                                                , count = Count + 1
```
id: Shouldn't it be `Count + Offset - End`?

src/brod_consumer.erl
```diff
@@ -442,17 +476,50 @@ handle_reset_offset(#state{offset_reset_policy = Policy} = State, _Error) ->
...
-handle_ack([H | Offsets], Offset) ->
-  case H =< Offset of
-    true  -> handle_ack(Offsets, Offset);
-    false -> [H | Offsets]
+handle_ack(#pending_acks{ offsets_queue = ?EMPTY_QUEUE
+                        , queue_rear = {Begin, End}
+                        , count = Count} = PendingAcks, Offset) ->
+  if
+    Offset >= End ->
+      NewCount = Count - (End - Begin + 1),
```
id: why `+1`?

zmstone: because it's inclusive at both ends? i.e. it's a range of `[Begin, End]` not `[Begin, End)`

theorix: yes ,I used [Begin, End]

zmstone:
awesome work!

zmstone (approved):

id (approved):

id:
Bump version?

zmstone:
is it ok to include it in the next fix release ?

PR #173, opened 2017-03-08
Add SASL/Plain authentication

codesnik:
requires https://github.com/klarna/kafka_protocol/pull/25

Erlang newbie here, so I'm all ears for your suggestions. `maybe_sasl_auth` looks pretty ugly IMHO. Also SASL error handling/reporting could be better, probably. I have no idea how (and if) to test for SASL errors.

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -28,8 +28,10 @@ fi
 ipwithnetmask="$(ip -f inet addr show dev eth0 | awk '/inet / { print $2 }')"
 ipaddress="${ipwithnetmask%/*}"
 
-sed -r -i "s/^(advertised.listeners)=(.*)/\1=PLAINTEXT:\/\/$ipaddress:$PLAINTEXT_PORT,SSL:\/\/$ipaddress:$SSL_PORT/g" $prop_file
-sed -r -i "s/^(listeners)=(.*)/\1=PLAINTEXT:\/\/:$PLAINTEXT_PORT,SSL:\/\/:$SSL_PORT/g" $prop_file
+sed -r -i "s/^(advertised.listeners)=(.*)/\1=PLAINTEXT:\/\/$ipaddress:$PLAINTEXT_PORT,SSL:\/\/$ipaddress:$SSL_PORT,SASL_SSL:\/\/$ipaddress:$SASL_SSL_PORT/g" $prop_file
+sed -r -i "s/^(listeners)=(.*)/\1=PLAINTEXT:\/\/:$PLAINTEXT_PORT,SSL:\/\/:$SSL_PORT,SASL_SSL:\/\/:$SASL_SSL_PORT/g" $prop_file
+echo "sasl.enabled.mechanisms=PLAIN" >> $prop_file
 
-/opt/kafka/bin/kafka-server-start.sh $prop_file
+KAFKA_HEAP_OPTS="-Xmx1G -Xms1G -Djava.security.auth.login.config=/etc/kafka/jaas.conf" \
+  /opt/kafka/bin/kafka-server-start.sh $prop_file
 
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
 {deps, [ {supervisor3, "1.1.5"}
-       , {kafka_protocol, "0.7.8"}
+       , {kafka_protocol, "0.8.0"}
        ]}.
 {erl_opts, [warn_unused_vars,warn_shadow_vars,warn_unused_import,warn_obsolete_guard,debug_info]}.
--- a/src/brod.app.src
+++ b/src/brod.app.src
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Apache Kafka Erlang client library"},
-  {vsn,"2.2.16"},
+  {vsn,"2.3.0"},
   {registered,[]},
   {applications,[kernel,stdlib,ssl,kafka_protocol,supervisor3]},
   {env,[]},
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -179,6 +179,9 @@ start_client(BootstrapEndpoints, ClientId) ->
 %%       true | false | [{certfile, ...},{keyfile, ...},{cacertfile, ...}]
 %%       When true, brod will try to upgrade tcp connection to ssl using default
 %%       ssl options. List of ssl options implies ssl=true.
+%%     sasl (optional, default=undefined)
+%%       Credentials for SASL/Plain authentication.
+%%       {plain, "username", "password"}
 %%     connect_timeout (optional, default=5000)
 %%       Timeout when trying to connect to one endpoint.
 %%     request_timeout (optional, default=120000, constraint: >= 1000)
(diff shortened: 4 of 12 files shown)


src/brod_sock.erl
```diff
@@ -179,6 +179,9 @@ init(Parent, Host, Port, ClientId, Options) ->
       proc_lib:init_ack(Parent, {ok, self()}),
       ReqTimeout = get_request_timeout(Options),
       ok = send_assert_max_req_age(self(), ReqTimeout),
+      ok = maybe_sasl_auth(State#state.sock, State#state.mod, ClientId, Timeout,
```
zmstone: move this up to right after line 178

src/brod_sock.erl
```diff
@@ -203,6 +206,38 @@ maybe_upgrade_to_ssl(Sock, _Mod = ssl, SslOpts = [_|_], Timeout) ->
 maybe_upgrade_to_ssl(Sock, _Mod, _SslOpts, _Timeout) ->
   {ok, Sock}.
 
+maybe_sasl_auth(_Sock, _Mod, _ClientId, _Timeout, _SaslOpts = undefined) -> ok;
+maybe_sasl_auth(Sock, Mod, ClientId, Timeout,
```
zmstone, line 209: We should match on Mod `ssl` here to only allow sasl when TLS is in place.

zmstone: and log a warning message saying that brod has refused to enable sasl on plaintext socket.

zmstone:
Thanks a lot!
It's time to start a collective dev branch for 2.3.
Please update the merge target to `2.3-dev` instead.

src/brod_sock.erl
```diff
@@ -203,6 +206,38 @@ maybe_upgrade_to_ssl(Sock, _Mod = ssl, SslOpts = [_|_], Timeout) ->
...
+  HandshakeRequestBin = kpro:encode_request(ClientId, 0, HandshakeRequest),
+  ok = Mod:send(Sock, HandshakeRequestBin),
+  {ok, <<Len:32>>} = Mod:recv(Sock, 4, Timeout),
+  {ok, HandshakeResponseBin} = Mod:recv(Sock, Len, Timeout),
+  {[ #kpro_Response{ responseMessage = #kpro_SaslHandshakeResponse{
+                                          errorCode = ErrorCode }}],
+    <<>>} = kpro:decode_response(<<Len:32, HandshakeResponseBin/binary>>),
+  case ErrorCode of
+    no_error ->
+      SaslMessage = list_to_binary([0, unicode:characters_to_binary(SaslUser),
```
zmstone: no need to construct binary.
the second arg of ssl:send/2 allows `iodata() `
http://erlang.org/doc/man/ssl.html#send-2

codesnik: yeah, I know, but I need `byte_size` too.

rebar.config
```diff
@@ -1,4 +1,4 @@
 {deps, [ {supervisor3, "1.1.5"}
-       , {kafka_protocol, "0.7.8"}
+       , {kafka_protocol, "0.7.9"}
```
id: 0.8.0

src/brod.app.src
```diff
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Apache Kafka Erlang client library"},
-  {vsn,"2.2.16"},
+  {vsn,"2.2.17"},
```
id: 2.3.0

src/brod_sock.erl
zmstone: true.
please move it to a help function instead.
e.g. `make_sasl_message(User, Password) -> binary()`

src/brod_sock.erl
codesnik: but kafka happily works with PLAINTEXT_SASL listeners. Of course, using plain SASL without SSL in production is wrong on so many levels, but it's useful in debugging. In fact, that's how I did most development. So, maybe we should leave it to ops guys to decide, what do you think?

I'd like to log a warning when *kafka* refuses to accept SASL handshake (with `'UnsupportedSaslMechanism'` or `'IllegalSaslState'` , but I'm not sure where that piece should go. I'm just doing `exit({sasl_auth_error, ...})` for that.

id: I also think we should leave this for Kafka folks.

id: Meaning, clients should not block api which is totally legit from the server point of view.

zmstone: ok.
@codesnik no need to log warning, exit with a reason should generate a informative log.

src/brod.erl
```diff
@@ -179,6 +179,9 @@ start_client(BootstrapEndpoints, ClientId) ->
 %%       true | false | [{certfile, ...},{keyfile, ...},{cacertfile, ...}]
 %%       When true, brod will try to upgrade tcp connection to ssl using default
 %%       ssl options. List of ssl options implies ssl=true.
+%%     sasl (optional, default=undefined)
+%%       Credentials for SASL/Plain authentication.
+%%       {"PLAIN", "username", "password"}
```
codesnik: btw, is this ok? I pass "PLAIN" there in case someone would want GSSAPI for real, in future. But maybe using optlist here is more idiomatic, idk.

docker/docker-entrypoint.sh
```diff
@@ -28,8 +28,10 @@ fi
...
 
-sed -r -i "s/^(advertised.listeners)=(.*)/\1=PLAINTEXT:\/\/$ipaddress:$PLAINTEXT_PORT,SSL:\/\/$ipaddress:$SSL_PORT/g" $prop_file
-sed -r -i "s/^(listeners)=(.*)/\1=PLAINTEXT:\/\/:$PLAINTEXT_PORT,SSL:\/\/:$SSL_PORT/g" $prop_file
+sed -r -i "s/^(advertised.listeners)=(.*)/\1=PLAINTEXT:\/\/$ipaddress:$PLAINTEXT_PORT,SSL:\/\/$ipaddress:$SSL_PORT,SASL_SSL:\/\/$ipaddress:$SASL_SSL_PORT/g" $prop_file
+sed -r -i "s/^(listeners)=(.*)/\1=PLAINTEXT:\/\/:$PLAINTEXT_PORT,SSL:\/\/:$SSL_PORT,SASL_SSL:\/\/:$SASL_SSL_PORT/g" $prop_file
+echo "sasl.enabled.mechanisms=PLAIN" >> $prop_file
 
-/opt/kafka/bin/kafka-server-start.sh $prop_file
+KAFKA_HEAP_OPTS="-Xmx1G -Xms1G -Djava.security.auth.login.config=/etc/kafka/jaas.conf" \
+  /opt/kafka/bin/kafka-server-start.sh $prop_file
```
codesnik, line 36: passing -Dopts to java daemons is always weird.

src/brod.erl
zmstone: what would be the input look like for GSSAPI ?
in general, `{Tag, Data}` tuple is the most straightforward extendable schema.

codesnik:
So then, maybe, plain or 'PLAIN' would be better?

zmstone:
yes, go with plain (not 'PLAIN')

codesnik:
I forgot to update versions in Makefile. Now all's fine.

zmstone (approved):

zmstone:
awesome work! thanks a lot!

id (approved):

PR #174, opened 2017-03-10
Info on how to deal with connection errors

kevinbader:
I'm not sure whether this is a question or an issue. In any case, I think it's worth documenting the behavior/best-practice, so here goes the pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,14 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
 ok = brod:start_client([{"localhost", 9092}], brod_client_1, ClientConfig).
 ```
 
+## Automatic handling of connection errors
+
+You may use a supervisor to start, e.g., a group subscriber by calling `brod:start_link_group_subscriber` in the module's `start_link` method. In the same method, you should make sure the respective brod client instance is running already using `brod:start_link_client`. The linking will cause your supervisor to go down in case the brod client fails (so you should put the starting supervisor under supervision itself), but it will allow for a clean restart of both client and subscriber.
+
+One thing to note is that the brod client doesn't die in case the connection goes down. ~~By default, it is configured to reconnect to the brokers automatically~~ AUTOMATIC RECONNECT DOESN'T WORK THAT WAY.... WHY?
+
+Another thing worth mentioning is that the subscriber will fail very quickly in case the brod client can't connect, which causes the supervisor to give up and potentially bring down the VM. Consequently one has to implement an exponential back-off strategy, or at least a delay, for the call to `brod:start_link_group_subscriber`. IS THERE A BETTER WAY?
+
 # Producers
 
 ## Auto start producer with default producer config
--- a/README.md
+++ b/README.md
@@ -119,13 +119,61 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
 ok = brod:start_client([{"localhost", 9092}], brod_client_1, ClientConfig).
 ```
 
-## Automatic handling of connection errors
+## Handling of connection errors
 
-You may use a supervisor to start, e.g., a group subscriber by calling `brod:start_link_group_subscriber` in the module's `start_link` method. In the same method, you should make sure the respective brod client instance is running already using `brod:start_link_client`. The linking will cause your supervisor to go down in case the brod client fails (so you should put the starting supervisor under supervision itself), but it will allow for a clean restart of both client and subscriber.
+Usually you would have a supervisor that starts your worker module. In the worker's `start_link` method, you should do two things:
 
-One thing to note is that the brod client doesn't die in case the connection goes down. ~~By default, it is configured to reconnect to the brokers automatically~~ AUTOMATIC RECONNECT DOESN'T WORK THAT WAY.... WHY?
+1. Make sure the Brod client you're going to use is running by calling `brod:start_client`. That will start the client under Brod's supervisor, if a client with the given name does not already exist. Afterwards, Brod will automatically recover from connection issues.
+1. Start, e.g., a group subscriber by calling `brod:start_link_group_subscriber`, and return the pid for your supervisor to monitor.
 
-Another thing worth mentioning is that the subscriber will fail very quickly in case the brod client can't connect, which causes the supervisor to give up and potentially bring down the VM. Consequently one has to implement an exponential back-off strategy, or at least a delay, for the call to `brod:start_link_group_subscriber`. IS THERE A BETTER WAY?
+(Note that using `brod:start_link_client` to link the client to your supervisor is deprecated. Using it, a connection loss won't trigger a reconnect, even though the process linked to stays alive.)
+
+Also, the subscriber will fail very quickly in case the brod client can't connect, which causes the supervisor to give up and potentially bring down the VM. In order to mitigate this, an exponential back-off strategy is needed, or at least a delay, between calls to `brod:start_link_group_subscriber`. In Elixir this may look like this:
+
+```Elixir
+defmodule MyGroupSubscriber do
+  @doc """
+  Makes sure the Brod client is running and starts the group subscriber.
+  """
+  def start_link do
+    brod_client_id = :my_brod_client
+    :ok = :brod.start_client([localhost: 9092], brod_client_id)
+    safely_start_group_subscriber(brod_client_id, "my-consumer-group", ["some-topic"])
+  end
+
+  # Starts the group subscriber, using exponential back-off if the brod client is down.
+  defp safely_start_group_subscriber(
+      brod_client_id, consumer_group_id, topics,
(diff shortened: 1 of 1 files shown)


zmstone (commented):
Thank you!
some comments:
1. start_link_client is deprecated, should call start_client instead. i.e let brod_client run under brod_sup.
2. when brod_client is down it usually means the whole kafka cluster is down, in this sense, taking down the vm is not completely unreasonable. nonetheless, brod should never kill a vm, brod_sup will try restart the client after a delay.
3. you may use supervisor3 to start you group subscriber(no exponential delay support though)

we can add a subscriber supervisor under brod_sup and add a start_subscriber API, not sure if thats really helpful though.

kevinbader:
Thank you for you feedback!

ad 2.) I think that even if the whole Kafka cluster is down, the application shouldn't necessarily go down with it. If all components would act this way, all services would have to be restarted after Kafka dies (for whatever reason). I'd rather have them recover automatically :)

ad API, do you think it would make sense to add a back-off callback, or a delay of some kind, to the (group?) config that gets passed to `start_link_group_subscriber`?

README.md
```diff
@@ -119,6 +119,62 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
 ok = brod:start_client([{"localhost", 9092}], brod_client_1, ClientConfig).
 ```
 
+## Handling of connection errors
```
zmstone, line 122: This section looks like consumer specific. perhaps move it down to the `Consumers` section after Group Subscriber ?

README.md
```diff
@@ -119,6 +119,62 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
...
+
+Usually you would have a supervisor that starts your worker module. In the worker's `start_link` method, you should do two things:
+
+1. Make sure the Brod client you're going to use is running by calling `brod:start_client`. That will start the client under Brod's supervisor, if a client with the given name does not already exist. Afterwards, Brod will automatically recover from connection issues.
```
zmstone, line 126: This part is somewhat duplicated to the 'Start brod client on demand' section.
If you think that's not detailed enough there, please help to make it more clear.

README.md
```diff
@@ -119,6 +119,62 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
 ok = brod:start_client([{"localhost", 9092}], brod_client_1, ClientConfig).
 ```
 
+## Handling of connection errors
+
+Usually you would have a supervisor that starts your worker module. In the worker's `start_link` method, you should do two things:
```
zmstone, line 124: Perhaps better to have it added to 'Group subscriber (brod_group_subscriber)' section.
Also, it should be 'starts your worker process' not worker module.

README.md
```diff
@@ -119,6 +119,62 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
...
+
+1. Make sure the Brod client you're going to use is running by calling `brod:start_client`. That will start the client under Brod's supervisor, if a client with the given name does not already exist. Afterwards, Brod will automatically recover from connection issues.
+1. Start, e.g., a group subscriber by calling `brod:start_link_group_subscriber`, and return the pid for your supervisor to monitor.
+
+(Note that using `brod:start_link_client` to link the client to your supervisor is deprecated. Using it, a connection loss won't trigger a reconnect, even though the process linked to stays alive.)
```
zmstone, line 129: I'm not sure what you have observed. but here is how `brod_client` handles connection restart.
1. When `brod_client` start/restart, it'll try to fetch the broker metadata by looping through all the given endpoints, if ALL fail, it gives up and crash, later `brod_sup` should restart it after a configurable delay.
2. A metadata connection is established to fetch metadata, usually after 10 minutes of idel, kafka will close it, `brod_client` will be 'lazy' on it: only start it again when there is a need to fetch metadata
3. A set of payload connections will be established to produce/fetch messages. If a connection is down, there will be a configurable 'cool-down' seconds of delay before allowed to re-establish the connection. However, the re-establish is also 'lazy', it is triggered either by producer trying to send a produce request, or a consumer trying to send a fetch request.

start_link_client is deprecated, however it doesn't imply any difference in brod_client's behaviour when it comes to socket error handling.

README.md
```diff
@@ -119,6 +119,62 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
...
+Also, the subscriber will fail very quickly in case the brod client can't connect, which causes the supervisor to give up and potentially bring down the VM. In order to mitigate this, an exponential back-off strategy is needed, or at least a delay, between calls to `brod:start_link_group_subscriber`. In Elixir this may look like this:
+
+```Elixir
+defmodule MyGroupSubscriber do
+  @doc """
+  Makes sure the Brod client is running and starts the group subscriber.
+  """
+  def start_link do
+    brod_client_id = :my_brod_client
+    :ok = :brod.start_client([localhost: 9092], brod_client_id)
```
zmstone, line 140: It's better to start the client in MyGroupSubscriber's supervisor instead.

README.md
```diff
@@ -119,6 +119,62 @@ ClientConfig = [{reconnect_cool_down_seconds, 10}],
...
+    safely_start_group_subscriber(brod_client_id, "my-consumer-group", ["some-topic"])
+  end
+
+  # Starts the group subscriber, using exponential back-off if the brod client is down.
+  defp safely_start_group_subscriber(
+      brod_client_id, consumer_group_id, topics,
+      restart_delay_ms \\ 100, max_restart_delay_ms \\ 10_000
+  ) do
+    result = start_group_subscriber(brod_client_id, consumer_group_id, topics)
```
zmstone, line 149: It would be easier to simply match `{:ok, _}` here and 'let it crash' if kafka cluster is down.
Delayed retry can be implemented in supervisor, here is an example:
https://github.com/klarna/kastlex/blob/master/lib/kastlex/collectors_sup.ex#L25

Or you can implement a genserver to monitor group subscriber and restart it after
some exponential delay (i.e. not only handle the start_link errors),  but also handle
crashes after it is successfully started.

README.md
kevinbader: Hm. When using `start_link_client`, I observed one new process below my supervisor, with three children attached to it. When killing Kafka, one of those children died with it and didn't come back; consequently `start_link_group_subscriber` kept failing (I've let it retry every 10 seconds for a few minutes). When using `start_client` the client reconnected after restarting Kafka as expected, so I've assumed differences, but to be honest I didn't go through the code.

README.md
kevinbader: You're right, that's probably cleaner.

README.md
kevinbader: Seems I'm not the only one with the problem/idea:
http://stackoverflow.com/questions/3785738/supervisors-with-backoff

I'd like the idea of 'let it crash' in general; however, I want the exponential back-off only in the case the group subscriber startup fails. Any error later on should not be handled in a special way.

kevinbader:
Thank you for your thorough review! I'll test different ways of doing this; afterwards I'll revise or drop this pull request. Maybe it makes more sense to add a wiki page instead, we'll see.

README.md
kevinbader: You've mentioned that `brod_sup` restarts the client if connecting fails. But `brod:start_link_client` doesn't involve `brod_sup`, so that's why recovering fails, right?

README.md
kevinbader: And thanks for the link to supervisor3, I haven't noticed that before

README.md
zmstone: You are right.

> Using it, a connection loss won't trigger a reconnect.

This made me thought that you meant ANY connection loss won't trigger a reconnect.
But the fact is `brod_client` only crashes when ALL of the cluster members are unreachable.

>  Even though the process linked to stays alive

I don't think it's necessary to state the obvious because
ANY `start_link` caller should expect the returned pid to exit.
So it should either trap exit to receive an 'EXIT' message or
unlink then monitor it to receive a 'DOWN' message, and finally
apply whatever restart strategy fits the best to their use case.

zmstone: It'd be nice to be informative, but better to keep README as brief as an overview.
It's perhaps enough to just describe under what circumstance `brod_client` will crash (in the #Client section).
Better to move detailed descriptions to a WIKI page instead.

zmstone:
@kevinbader Thank you for your contribution.
Discussion like this helps us to identify the missing info in README.
It'd be nice to have below updates to README.
1. The #Client section updated with info about when `brod_client` will crash. And how the socket errors are handled in general.
2. To be more clear in the #Consumer section, that group subscribers are not started under `brod_sup` and the callers process (or supervisor) should take care of restarts.

A new wiki page for your backoff start demo is welcome.
I'm not sure if you can create a wiki here, please send me a snippet,
i'll add the page, then we can revise it together.

PR #192, opened 2017-04-04
Ensure all sent requests put back to buffer when new socket found

### LINKED ISSUE #189: Processes calling brod:produce_sync sometimes block forever
Hello again,

We've been investigating issues where our number of process would slowly increase. This always happens after one of the kafka server nodes becomes unavailable (for even a very short time), leading to kafka push errors (NotLeaderForPartition) during the unavailability. Once the node is back, things look like they are back to normal, except that we have the steady increase in blocked processes.

All the blocked processes are stuck in brod_producer:produce line 150, with the following erlang stacktraces:
```
 {{current_stacktrace,[{brod_producer,produce,3,[{file,"src/brod_producer.erl"},{line,150}]},
                       {brod,produce_sync,5,[{file,"src/brod.erl"},{line,340}]},
                       [**our process callstack**] ]},
```
This corresponds to a receive, waiting for #brod_produce_reply with result=brod_produce_req_buffered.

I added some debug to brod, waited for the issue to reappear, and I think I've cornered why they block forever:
The processes call: brod:produce_sync -> brod:produce -> brod_producer:produce -> Pid ! {produce, CallRef, Key, Value} & receive waiting for #brod_produce_reply with result=brod_produce_req_buffered.

However, in brod_producer_buffer:maybe_buffer, if BufferCount >= BufferLimit we will never call reply_buffered(Req), which means that the process will never receive brod_produce_req_buffered.

I presume the temporary push errors lead to the buffer filling up, which is why we don't usually see the issue, but I'm not sure :p.

Does my analysis hold up? Or am I misunderstanding what is happening?

Edit: Reproduced with brod 2.3.1 (originally we found the bug with 2.2.12) and kafka 0.10.1.1

zmstone:
Fixes #189


src/brod_producer.erl
```diff
@@ -434,15 +442,15 @@ maybe_demonitor(Mref) ->
 %% @private
 schedule_retry(#state{buffer = Buffer} = State, Reason) ->
   {ok, NewBuffer} = brod_producer_buffer:nack_all(Buffer, Reason),
```
id: I feel it will be more clear if we call `nack_all` explicitly where it belongs and have only one version of `schedule_retry` so that the function does exactly what the name implies.

src/brod_producer.erl
```diff
@@ -359,19 +347,36 @@ log_discarded_corr_id(CorrIdReceived, CorrIdExpected) ->
...
-handle_produce(CallRef, Key, Value, #state{buffer = Buffer} = State) ->
+%% @private
+handle_produce(CallRef, Key, Value,
+               #state{retry_tref = Ref} = State) when is_reference(Ref) ->
+  %% pending on retry, add to buffer regardless of socket state
+  do_handle_produce(CallRef, Key, Value, State);
+handle_produce(CallRef, Key, Value,
+               #state{sock_pid = Pid} = State) when is_pid(Pid) ->
+  %% Socket is alive, add to buffer, and try send produce request
+  do_handle_produce(CallRef, Key, Value, State);
```
id, line 364: why treat 2 clauses above as separate ones?

id: never mind, I see it

id (approved):

id (approved):

PR #193, opened 2017-04-06
Smart direct fetch

zmstone:


src/brod_utils.erl
```diff
@@ -172,8 +148,49 @@ assert_topic(Topic) ->
...
+   is_record(M, kpro_Message) andalso
+   M#kpro_Message.offset >= BeginOffset].
+
+
 %%%_* Internal Functions =======================================================
 
+%% @private Convert a `kpro_Message' to a `kafka_message'.
+-spec kafka_message(#kpro_Message{})     -> #kafka_message{};
+                   (?incomplete_message) -> ?incomplete_message.
+kafka_message(#kpro_Message{ offset     = Offset
```
id: Will it work if we just `setelement(1, Msg, kafka_message)`?

zmstone: no, field `messageSize` from kpro record is not included in `#kafka_message{}`.
and there might be more fields added in the next kafka protocol version which are useless for consumers.
Or perhaps in 3.0, we can simply expose kpro records.

id (approved):

PR #194, opened 2017-04-19
Adapt brod for new version of kafka_protocol

codesnik:
work in progress. Just to see how API would feel when used.
Tests are failing on my machine, something's definitely broken in https://github.com/klarna/kafka_protocol/pull/28


zmstone:
covered in #219

PR #208, opened 2017-05-19
SASL GSSAPI support

ElMaxo:
I have implemented experimental SASL GSSAPI auth mechanism support. To enable this - add `{sasl, {kerberos, Keytab, Principal}}` to client config. Please, let me know if you found any bugs. I want to add automatic ticket renew in future

### DIFF
--- a/c_src/Makefile
+++ b/c_src/Makefile
@@ -1,0 +1,74 @@
+# Based on c_src.mk from erlang.mk by Loic Hoguin <essen@ninenines.eu>
+
+CURDIR := $(shell pwd)
+BASEDIR := $(abspath $(CURDIR)/..)
+
+PROJECT ?= $(notdir $(BASEDIR))
+PROJECT := $(strip $(PROJECT))
+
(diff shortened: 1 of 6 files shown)


id:
@ElMaxo nice, thanks! We'll take a look.

ElMaxo:
To build sasl_auth.c libsasl2 (cyrus sasl) should be installed with headers

zmstone:
@ElMaxo Thank you for your contribution.
This adds more compile time dependency for the users who don't need this new auth support.
We'd like to have this supported in a more flexible way:
1. Put the nif code to a separate Erlang project.
2. Add a `{auth_callback_module, ModuleName}` client config (and maybe introduce a `-behaviour` module to define the callback specs)
3. Update README with info about how to enable SASL GSSAPI auth.

ElMaxo:
@zmstone ok, I'll do it in some days and write you back

zmstone:
Covered by #210

PR #216, opened 2017-06-07
Support sasl-plain username password in text file

### LINKED ISSUE #186: do not store sasl password in process state
the password might get dumped to crash logs.

zmstone:
fixes #186

### DIFF
--- a/src/brod_utils.erl
+++ b/src/brod_utils.erl
@@ -42,6 +42,8 @@
         , fetch_offsets/5
         , map_messages/2
         , fetch/4
+        , init_sasl_opt/1
+        , get_sasl_opt/1
         ]).
 
 -include("brod_int.hrl").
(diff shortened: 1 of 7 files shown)


src/brod_utils.erl
```diff
@@ -210,9 +212,55 @@ fetch(SockPid, ReqFun, MaxBytes, Offset) ->
...
+%% the plain text being dumped to crash logs
+%% @end
+-spec init_sasl_opt(client_config()) -> client_config().
+init_sasl_opt(Config) ->
+  case proplists:get_value(sasl, Config) of
+    {plain, User, Pass} when is_list(Pass) orelse is_binary(Pass) ->
+      replace_prop(sasl, {plain, User, fun() -> Pass end}, Config);
+    {plain, File} ->
+      {User, Pass} = read_sasl_file(File),
+      replace_prop(sasl, {plain, User, fun() -> Pass end}, Config);
```
id: lists:keystore?

src/brod_utils.erl
```diff
@@ -210,9 +212,55 @@ fetch(SockPid, ReqFun, MaxBytes, Offset) ->
       end
   end.
 
+%% @doc Get sasl options from client config.
+-spec get_sasl_opt(client_config()) -> sasl_opt().
+get_sasl_opt(Config) ->
+  case proplists:get_value(sasl, Config) of
+    {plain, User, PassFun} when is_function(PassFun) ->
+      {plain, User, PassFun()};
+    {plain, File} ->
```
id, line 221: in which situation this case will trigger?

zmstone: brod-cli. or when calling `brod:connect_leader/4` directly or indirectly e.g. from `brod:fetch/8`

id: get_sasl_opt and init_sasl_opt are mostly identical. Perhaps leave just this one?

zmstone: init is to hide password in anonymous fun
get is to evaluate the anonymous fun

src/brod_utils.erl
zmstone: we use `proplists:` APIs everywhere for client config

src/brod_utils.erl
id: if you call `brod:connect_leader/4 ` directly then you don't have a fun to evaluate, or?

id: nvm, then it will pass User/Pass as is.

id (approved):

PR #219, opened 2017-06-14
Make it work with kafka_protocol 1.0.0

zmstone:


Makefile
```diff
@@ -10,7 +10,7 @@ ERLC_OPTS = -Werror +warn_unused_vars +warn_shadow_vars +warn_unused_import +war
 TEST_ERLC_OPTS = -Werror +warn_unused_vars +warn_shadow_vars +warn_unused_import +warn_obsolete_guard +debug_info
 
 dep_supervisor3_commit = 1.1.5
-dep_kafka_protocol_commit = 0.9.2
+dep_kafka_protocol = git https://github.com/klarna/kafka_protocol.git new-kafka-api
```
id: use same version in rebar.config

id (approved):
Great work.

zmstone:
merged and resolved conflicts from contribution by @onno-vos-dev 
will have to fix kafka_protocol commit to 1.0.1 before merge.

after merge, won't tag 3.0.0 for now, will have add more features (such as API version picks) before tagging

src/brod_group_subscriber.erl
```diff
@@ -93,43 +96,46 @@
 %% default value -1 (latest) is used.
 %
 % commented out as it's an optional callback
-%-callback get_committed_offsets(group_id(), [{topic(), partition()}],
-%                                cb_state()) ->
-%            {ok, [{{topic(), partition()}, offset()}], cb_state()}.
-
-
+%-callback get_committed_offsets(brod:group_id(),
```
id, line 99: We have `-optional_callbacks` attribute since OTP18 but I guess it's a bit too early to deprecate 17. Just a note.

src/brod_utils.erl
```diff
@@ -237,8 +277,195 @@ init_sasl_opt(Config) ->
...
+
+%% @doc List all groups in the given cluster.
+%% NOTE: Exception if failed against any of the coordinator brokers.
+%% @end
+-spec list_all_groups([endpoint()], sock_opts()) ->
+        [{endpoint(), [brod:cg()] | {error, any()}}].
+list_all_groups(Endpoints, Options) ->
+  {ok, Metadata} = get_metadata(Endpoints, [], Options),
+  Brokers0 = kf(brokers, Metadata),
+  Brokers = [{binary_to_list(kf(host, B)), kf(port, B)} || B <- Brokers0],
```
id, line 362: This is kinda annoying, perhaps we should allow hostname to be binary as well as list in brod_sock? Not necessarily in this version.

zmstone: I tried. However Hostname being a string() is assertively used in many places.
It is also used as ETS keys.
Doesn't gain much to support `binary() | string()` spec everywhere.

I'd prefer to support a `raw_hostname() :: binary() | string()` for public APIs
and ensure `hostname() :: string()` everywhere internally.

src/brod_client.erl
```diff
@@ -843,20 +869,20 @@ do_start_consumer(TopicName, ConsumerConfig, State) ->
       end,
   do_on_valid_topic(TopicName, State, F).
 
+%% @private
 do_on_valid_topic(TopicName, State, F) ->
```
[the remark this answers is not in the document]
zmstone: rename to `ensure_partition_workers`
was planing to use it as a more generic function, but so far, it is only to start workers.

src/brod_group_coordinator.erl
```diff
@@ -512,37 +517,37 @@ join_group(#state{ groupId                       = GroupId
...
-    } = JoinRsp,
+  #kpro_rsp{ tag = join_group_response
+           , vsn = Vsn
+           , msg = RspBody
+           } = send_sync(SockPid, Req, SessionTimeout),
+  ?ESCALATE_EC(kpro:find(error_code, RspBody)),
+  GenerationId = kpro:find(generation_id,RspBody),
+  LeaderId = kpro:find(leader_id, RspBody),
+  MemberId = kpro:find(member_id, RspBody),
+  Members = kpro:find(members, RspBody), %% assertive match
```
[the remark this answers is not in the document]
zmstone: removed.

id (approved):
🥇

PR #228, opened 2017-09-04
add erlang 19, 20 for test

getong:


id (approved):

test/brod_group_subscriber_SUITE.erl
```diff
@@ -57,6 +58,14 @@
 -define(GROUP_ID, list_to_binary(atom_to_list(?MODULE))).
 -define(config(Name), proplists:get_value(Name, Config)).
 
+-ifdef(old_rand_uniform).
+rand_uniform(Max) ->
```
zmstone: this new function is not used. (why export?)

test/brod_group_subscriber_SUITE.erl
```diff
@@ -36,6 +36,7 @@
 -export([ init/2
         , get_committed_offsets/3
         , handle_message/4
+		, rand_uniform/1
```
zmstone: is this tab? we use spaces.

rebar.config
```diff
@@ -1,4 +1,11 @@
...
        , {kafka_protocol, "1.0.0"}
        ]}.
-{erl_opts, [warn_unused_vars,warn_shadow_vars,warn_unused_import,warn_obsolete_guard,debug_info]}.
+{erl_opts, [
+		   warn_unused_vars,
+		   warn_shadow_vars,
+		   warn_unused_import,
+		   warn_obsolete_guard,
+		   debug_info,
+		   {platform_define, "17|18|19", old_rand_uniform}
```
zmstone: if this is only for the deprecated `crypto:rand_uniform` I'd say don't add compile flags since that is used only in tests. simply take the system time's micro part is good enough for testing.

getong:
@zmstone  The reason why test failed seems to be the network, and in the session `make t` fail.

zmstone:
@getong I'm for not to introduce compile flags for this fix.
as `{platform_define, "^(17|18|19)", old_rand_uniform}` this is only for rebar,
we'll otherwise have to do similar for erlang.mk
(brod supports builds for rebar rebar3 and erlang.mk)

This only affects test code (to randomize a partition number).
It's fine to have some less random-ish numbers for this case.
e.g. `rand(Max) -> {_, _, Micro} = os:timestamp(), Micro rem Max.`

getong:
@zmstone  I will modify the code as you suggested.

PR #229, opened 2017-09-07
Add option to allow topic subscriber handling Kafka message or message_set

jfacorro:


src/brod_topic_subscriber.erl
```diff
@@ -299,6 +330,23 @@ subscribe_partition(Client, Topic, Consumer) ->
       end
   end.
 
+handle_message_set(Partition, MessageSet, State) ->
+  #state{cb_fun = CbFun, cb_state = CbState} = State,
+  ErlRef = erlang:make_ref(),
+  AckRef = {Partition, ErlRef},
```
zmstone: `AckRef` should be `{partition(), offset()}`
Here it should be the offset of the last message in the received message set
(you may safely call `lists:last/1` because message-set is never empty)

Also, there is no need to evaluate `lists:last/1` when `AckNow` is not true

zmstone:
@jfacorro very nice!

Could you add this feature to `brod_group_subscriber` too?
And please bump version number to 3.1.0 (both in Makefile and .app.src)

src/brod_topic_subscriber.erl
jfacorro: 👍

zmstone:
please also update changelog.md

zmstone (approved):

zmstone:
will squash merge after approval from @id

id (approved):
Good stuff, thanks!

PR #246, opened 2017-12-02
Introduce roundrobin_v2 group protocol

### LINKED ISSUE #241: Constant lag of one message on every partition
Hi!
I noticed that the `kafka-consumer-groups.sh` reports a one message lag for every partition when using a brod consumer.
`kafka-consumer-groups.sh --new-consumer --bootstrap-server kafkahost:9092 --describe --group my_consumer_group` reports

GROUP | TOPIC | PARTITION | CURRENT-OFFSET | LOG-END-OFFSET | LAG | OWNER
--------|--------|------------|--------------------|--------------------|------|--------
my_consumer_group | test_topic | 0 | 3 | 4 | 1 |nonode@nohost/<0.1036.0>_/172.26.0.1

with one partition but if the topic has more partition the lag is one for each.

I believe the committed offset does not follow the kafka consumer documentation at https://kafka.apache.org/0110/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html

> Note: The committed offset should always be the offset of the next message that your application will read. Thus, when calling commitSync(offsets) you should add one to the offset of the last message processed. 

I think brod doesn't add one when committing but at the same time when the consumer starts it adds one to received offset from kafka so it works good and don't skip messages.

This issue doesn't create incorrect behaviour, however we use a lag monitoring tool and only the brod consumer reports a constant lag. Other clients (java, scala) follow the kafka consumer recommendation.

Could you confirm my theory and maybe fix this if you agree?

zmstone:
Fixes #241 

### Changes

Prior to this change, brod's 'roundrobin' group protocol commits
acknowledged offsets to kafka, then +1 as begin-offset when assign
to group members to continue after restart/rebalance.

Starting from this version, brod will commit 'next offset to fetch',
to be compliant with the protocol spec and other kafka client
implementations, as well as monitoring tools.

### Upgrade offsets committed to kafka

'roundrobin_v2' will upgrade old commits to new version by +1.
This is done by making use of the metadata committed together
with the offsets:
  'roundrobin' metadata should be a string containing the Erlang node
  name and the group coordinator pid. e.g. "myconsumer@myhost/<0,6,0>".
  'roundrogin_v2' uses the same but with a `"+1/"` prefix.

#### NOTE

Since it's a new protocol, it does not support rolling upgrade
from a running 'roundrobin' group. i.e. all 'roundrobin' members
will have to be stopped before starting 'roundrobin_v2' members.

Nonetheless, there is no risk even if 'roudrobin' and
'roundrobin_v2' members are started at the same time,
because kafka would not allow a member supporting different
protocol to join an active group.

It is however NOT OK for 'roundrobin' members to pick up
'roundrobin_v2' commits. i.e. downgrade is not supported.

### For consumer managed offsets:

For consumer managed offsets (offset_commit_policy = consumer_managed):
There is no way to smooth upgrade already committed offsets,
therefor we will keep commiting 'acked offsets' instead of 
'begin offset' until next major release.

### DIFF
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -941,15 +941,15 @@ resolve_begin_offsets([{Topic, Partition} | Rest], CommittedOffsets,
       {_, Offset} when IsConsumerManaged ->
         %% roundrobin_v2 is only for kafka commits
         %% for consumer managed offsets, it's still acked offsets
-        %% therefore we need to +1 as begin-offset
+        %% therefore we need to +1 as begin_offset
         Offset + 1;
       {_, Offset} ->
-        %% offsets committed to kafka is already begin offset
+        %% offsets committed to kafka is already begin_offset
         %% since the introduction of 'roundrobin_v2' protocol
         Offset;
       false  ->
         %% No commit history found
-        %% Should use default begin offset in consumer config
+        %% Should use default begin_offset in consumer config
         ?undef
     end,
   Assignment =
@@ -1100,8 +1100,7 @@ is_roundrobin_v1_commit(Metadata) ->
   end.
 
 %% @private Upgrade offset from old roundrobin protocol to new.
-%% old (before roundrobin_v2) brod commits acked offsets
-%% not the next offset (begin offset) to fetch
+%% old (before roundrobin_v2) brod commits acked offsets not begin_offset
 %% @end
 -spec maybe_upgrade_from_roundrobin_v1(brod:offset(), binary()) ->
         brod:offset().
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -941,7 +941,7 @@ resolve_begin_offsets([{Topic, Partition} | Rest], CommittedOffsets,
       {_, Offset} when IsConsumerManaged ->
         %% roundrobin_v2 is only for kafka commits
         %% for consumer managed offsets, it's still acked offsets
-        %% therefor we need to +1 as begin-offset
+        %% therefore we need to +1 as begin-offset
         Offset + 1;
       {_, Offset} ->
         %% offsets committed to kafka is already begin offset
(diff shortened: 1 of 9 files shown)


src/brod_group_coordinator.erl
```diff
@@ -911,44 +913,53 @@ get_committed_offsets(#state{ offset_commit_policy = commit_to_kafka_v2
...
+resolve_begin_offsets([{Topic, Partition} | Rest], CommittedOffsets,
+                      IsConsumerManaged) ->
+  BeginOffset =
     case lists:keyfind({Topic, Partition}, 1, CommittedOffsets) of
-      {_, Offset_} when is_integer(Offset_) ->
-        Offset_;
+      {_, Offset} when IsConsumerManaged ->
+        %% roundrobin_v2 is only for kafka commits
+        %% for consumer managed offsets, it's still acked offsets
+        %% therefor we need to +1 as begin-offset
```
id: therefore.

Also, how "consumer managed" flag is related to roundrobin_v2?

src/brod_group_coordinator.erl
```diff
@@ -1023,7 +1034,10 @@ log(#state{ groupId  = GroupId
 
 %% @private Make metata to be committed together with offsets.
 -spec make_offset_commit_metadata() -> binary().
-make_offset_commit_metadata() -> coordinator_id().
+make_offset_commit_metadata() ->
+  %% Use a '+1/' prefix as a commit from group member which supports
```
id, line 1038: why `+1` and not `roundrobin_v2`?

zmstone: mainly to keep it short. it is repeated for each and every offset commit.

src/brod_group_coordinator.erl
zmstone: `resolve_begin_offsets/3` is a function shared for `consumer_managed` offset commit policy and `commit_to_kafka_v2` policy.

id (approved):

PR #256, opened 2018-02-02
Restart subscriber from first unseen offset in case of consumer crash

dszoboszlay:
`brod_topic_subscriber` and `brod_group_subscriber` shall handle a crash of `brod_consumer` transparent to the user. However, they used to resubscribe from the last acked offset with the `brod_consumer` process after its restart. This is not correct, as this would lead to double delivery of messages that had been delivered, but hadn't been acked before the crash.

This change fixes this problem (see the new `t_consumer_crash` test cases that would fail without this patch), and also makes sure that after a resubscribe the prefetch window of the `brod_consumer` is reduced by the number of outstanding unacked messages. This is to prevent a rapidly crashing and restarting `brod_consumer` to flood a slow async subscriber with messages.

### DIFF
--- a/src/brod_consumer.erl
+++ b/src/brod_consumer.erl
@@ -124,7 +124,7 @@ start_link(ClientPid, Topic, Partition, Config) ->
 %%  sleep_timeout (optional, default = 1000 ms):
 %%     Allow consumer process to sleep this amout of ms if kafka replied
 %%     'empty' message-set.
-%%  prefetch_count (optional, default = 1):
+%%  prefetch_count (optional, default = 10):
 %%     The window size (number of messages) allowed to fetch-ahead.
 %%  begin_offset (optional, default = latest):
 %%     The offset from which to begin fetch requests.
@@ -160,6 +160,22 @@ stop(Pid) -> safe_gen_call(Pid, stop, infinity).
 %%     offset to a persistent storage, then start (or restart) with
 %%     last_processed_offset + 1 as the begin_offset to proceed.
 %%     By default, it fetches from the latest available offset.
+%%   acked_offset (optional, default = begin_offset - 1)
+%%     A subscriber may limit the prefetch window on startup with this
+%%     option. The consumer will assume offsets from acked_offset up
+%%     to but not including the effective begin_offset it calculates
+%%     will be acked in the future by the subscriber. It will reduce
+%%     the prefetch window with the number of such offsets.
+%%
+%%     This feature is most useful when the brod_consumer process
+%%     crashes, but the subscriber process survives and attempts to
+%%     subscribe again once the brod_consumer is restarted. In this
+%%     case the subscriber may still have some unacked messages
+%%     received from the crashed brod_conumser. It needs to restart
+%%     from the last offset + 1 it has seen, so it won't receive
+%%     duplicated messages, but also needs to reduce the initial
+%%     prefetch window, otherwise a frequently crashing brod_consumer
+%%     may flood it with messages.
 %% @end
 -spec subscribe(pid(), pid(), options()) -> ok | {error, any()}.
 subscribe(Pid, SubscriberPid, ConsumerOptions) ->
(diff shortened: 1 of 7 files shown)


src/brod_consumer.erl
```diff
@@ -160,6 +160,22 @@ stop(Pid) -> safe_gen_call(Pid, stop, infinity).
...
+%%     option. The consumer will assume offsets from acked_offset up
+%%     to but not including the effective begin_offset it calculates
+%%     will be acked in the future by the subscriber. It will reduce
+%%     the prefetch window with the number of such offsets.
+%%
+%%     This feature is most useful when the brod_consumer process
+%%     crashes, but the subscriber process survives and attempts to
+%%     subscribe again once the brod_consumer is restarted. In this
+%%     case the subscriber may still have some unacked messages
+%%     received from the crashed brod_conumser. It needs to restart
```
mikpe, line 174: Nit: typo, s/conumser/consumer/g

src/brod_consumer.erl
```diff
@@ -160,6 +160,22 @@ stop(Pid) -> safe_gen_call(Pid, stop, infinity).
...
+%%     from the last offset + 1 it has seen, so it won't receive
```
mikpe, line 175: "from the last offset + 1 it has seen" is ambiguous to me; does it mean "from one past the last offset it has seen" or does it mean (constraint-programming style) "from Y, where Y+1 is the last offset it has seen"?

src/brod_group_subscriber.erl
```diff
@@ -465,19 +465,26 @@ handle_message_set(MessageSet, State) ->
         erlang:error({bad_return_value,
                       {CbModule, handle_message, Unknown}})
     end,
+  LastOffset = (lists:last(Messages))#kafka_message.offset,
```
mikpe, line 468: We expect Messages to be a short list I hope?

src/brod_group_subscriber.erl
```diff
@@ -465,19 +465,26 @@ handle_message_set(MessageSet, State) ->
...
+      AckRef = {Topic, Partition, LastOffset},
+      handle_ack(AckRef, State2);
+    false -> State2
   end.
 
 handle_messages(_Topic, _Partition, [], State) ->
   State;
-handle_messages(Topic, Partition, [Msg | Rest], State) ->
+handle_messages(Topic, Partition, Messages, State) ->
+  LastOffset = (lists:last(Messages))#kafka_message.offset,
```
mikpe, line 481: Same here: is Messages expected to be a short list?

src/brod_group_subscriber.erl
```diff
@@ -543,23 +568,31 @@ subscribe_partitions(#state{ client    = Client
...
-      %% otherwise fetch from the assigned begin_offset
-      BeginOffset = case AckedOffset of
-                      ?undef        -> BeginOffset0;
-                      N when N >= 0 -> N + 1
-                    end,
       Options =
-        case BeginOffset =:= ?undef of
-          true  -> []; %% fetch from 'begin_offset' in consumer config
-          false -> [{begin_offset, BeginOffset}]
+        if BeginOffset =:= ?undef andalso
```
mikpe, line 579: The logic in this block looks duplicated elsewhere and might warrant being factored out into a helper function.

src/brod_topic_subscriber.erl
```diff
@@ -345,19 +359,26 @@ handle_message_set(MessageSet, State) ->
...
+      AckRef = {Partition, LastOffset},
+      handle_ack(AckRef, State2);
+    false -> State2
   end.
 
 handle_messages(_Partition, [], State) ->
   State;
-handle_messages(Partition, [Msg | Rest], State) ->
+handle_messages(Partition, Messages, State) ->
+  LastOffset = (lists:last(Messages))#kafka_message.offset,
```
mikpe, line 375: lists:last/1 again...

mikpe (approved):
LGTM.  Since I don't know this code well I'll leave it up to you if any of my comments need addressing or not.

src/brod_topic_subscriber.erl
```diff
@@ -85,6 +85,7 @@
         { partition     :: brod:partition()
         , consumer_pid  :: ?undef | pid() | {down, string(), any()}
         , consumer_mref :: ?undef | reference()
+        , begin_offset  :: ?undef | brod:offset()
```
zmstone, line 88: now that we allow special offset, it should be `brod:offset_time()` here.

src/brod_consumer.erl
```diff
@@ -570,49 +586,59 @@ send_fetch_request(#state{ begin_offset = BeginOffset
...
+-spec make_pending_acks(offset(), offset()) -> pending_acks().
+make_pending_acks(AckedOffset, BeginOffset) when
+    AckedOffset >= BeginOffset - 1 ->
+  #pending_acks{};
+make_pending_acks(AckedOffset, BeginOffset) ->
+  Begin = AckedOffset + 1,
+  End = BeginOffset - 1,
+  Count = End - Begin + 1,
+  Queue = queue:in({Begin, End}, queue:new()),
+  #pending_acks{offsets_queue = Queue, count = Count}.
```
zmstone, line 641: Kafka offsets may not be sequential, there can be gaps in compacted topics-partitions.
e.g. if message 1 has key 'a' and message 2, 3 both has key 'b', eventually message 2 will be compacted away leaving only message 1 and 3 on disk.

If we use `begin - acked` as the initial pre-fetch count,
`brod_consumer` may end up stuck with a too large made-up pending queue
and not fetching anything until the last message is acked.

e.g. if a message set has two messages `[offset = 1, offset = 99]`,
and if `brod_consumer` having a `prefetch_count = 10`,
it will not fetch anything until `offset = 99` is acked. 

To be accurate, the same offset segment queue (`queue:queue({Bgn::offset(), End::offset()})`)
should be tracked in `brod_xxx_subscriber`, and give this queue back to `brod_consumer` when re-subscribe.

I'm fine keeping current implementation.
I'd say this is a minor defect as 'fetch too few', which is better than 'fetch too many',
however it justifies a comment here when making up the initial pending queue.

zmstone:
let's step back and think again: when does a `brod_consumer` restart ?
- it may crash due to a bug: this may cause a lot re-delivery without this fix.
  but the right thing to do would be to fix the bug that causes it to restart.
- it received non-retriable error code from kafka, which will cause a delayed restart of `brod_consumer`. in this case, there is a `#kafka_fetch_error{}` delivered to the subscriber
  before `brod_consumer` restarts itself. maybe we should add offset in `#kafka_fetch_error{}`
  so the subscriber can make use of it?

zmstone:
fixed in #291

PR #257, opened 2018-02-12
Fix incorrect typespec

CJPoll:
brod:start_link_group_subscriber/8 has an incorrect typespec.

It declares `message | message_type` for the message type, but should be `message | message_set`.
This causes dialyzer to freak out in my client apllication. 

Ideally there would be a `message_type()` type that all typespecs could refer to, but this commit at least removes the error.


zmstone:
thanks for the PR.

PR #260, opened 2018-02-28
Prefetch bytes consumer config

zmstone:


id (approved):

src/brod_consumer.erl
```diff
@@ -197,7 +213,8 @@ init({ClientPid, Topic, Partition, Config}) ->
   MaxBytes = Cfg(max_bytes, ?DEFAULT_MAX_BYTES),
   MaxWaitTime = Cfg(max_wait_time, ?DEFAULT_MAX_WAIT_TIME),
   SleepTimeout = Cfg(sleep_timeout, ?DEFAULT_SLEEP_TIMEOUT),
-  PrefetchCount = erlang:max(Cfg(prefetch_count, ?DEFAULT_PREFETCH_COUNT), 1),
+  PrefetchCount = erlang:max(Cfg(prefetch_count, ?DEFAULT_PREFETCH_COUNT), 0),
```
viktoriafordos, line 216: What will happen if `PrefetchCount` is `0`?

zmstone: PrefetchCount was used as minimum size stat window size, wrongly forced to be 1.
Now PrefetchCount is only used to limit fetch ahead.
usage is: `PendingAcks.count > PrefetchCount` andalso `PendingAcks.bytes > PrefetchByets` -> then stop sending more.
`PendingAcks,count` and `bytes` are both initialised with `0`,
so the `0 > 0 -> false` result is as expected.

src/brod_consumer.erl
```diff
@@ -366,68 +384,65 @@ handle_message_set(#kafka_message_set{messages = Messages} = MsgSet,
...
+    queue:peek_r(NewPendingAcks#pending_acks.queue),
+  State1 = State0#state{ pending_acks = NewPendingAcks
+                       , begin_offset = LastOffset + 1
+                       },
+  State2 = maybe_shrink_max_bytes(State1, MsgSet#kafka_message_set.messages),
+  State = maybe_send_fetch_request(State2),
+  {noreply, State}.
 
-%% @private
+%% Add received offsets to offset range queue.
```
zmstone: it's no longer a range, will fix comment.

viktoriafordos (approved):

PR #265, opened 2018-04-12
resolve_offset doesn't exist anymore

stephanos:
Update the README to a function that exists. Honestly, not sure whether `brod_utils:fetch_offsets` or `brod:resolve_offset` would be right here but it's still better than a method that doesn't exist anymore.


zmstone:
thanks

PR #271, opened 2018-06-27
Add new produce API call, returning the assigned offset 

aronisstav:
Kafka responses for produce requests that are acknowledged include the offset that was assigned to the message. This PR adds an API call for retrieving this offset. 

It also includes a number of small fixes.

Feedback is very welcome. :-)

### DIFF
--- a/include/brod.hrl
+++ b/include/brod.hrl
@@ -44,7 +44,10 @@
                        , ref    :: reference()
                        }).
 
+-define(BROD_PRODUCE_UNKNOWN_OFFSET, -1).
+
 -record(brod_produce_reply, { call_ref :: brod:call_ref()
+                            , offset   :: brod:offset()
                             , result   :: brod:produce_result()
                             }).
 
--- a/include/brod.hrl
+++ b/include/brod.hrl
@@ -47,7 +47,7 @@
 -define(BROD_PRODUCE_UNKNOWN_OFFSET, -1).
 
 -record(brod_produce_reply, { call_ref :: brod:call_ref()
-                            , offset   :: brod:offset()
+                            , base_offset :: brod:offset()
                             , result   :: brod:produce_result()
                             }).
 
--- a/src/brod_producer_buffer.erl
+++ b/src/brod_producer_buffer.erl
@@ -149,8 +149,7 @@ ack(Buf, CorrId) ->
 ack(#buf{ onwire_count = OnWireCount
         , onwire       = [{CorrId, Reqs} | Rest]
         } = Buf, CorrId, BaseOffset) ->
-  Fold = fun(R, Offset) -> reply_acked(R, Offset) end,
-  _ = lists:foldl(Fold, BaseOffset, Reqs),
+  _ = lists:foldl(fun reply_acked/2, BaseOffset, Reqs),
   {ok, Buf#buf{ onwire_count = OnWireCount - 1
               , onwire       = Rest
               }};
--- a/src/brod_producer_buffer.erl
+++ b/src/brod_producer_buffer.erl
@@ -383,7 +383,7 @@ maybe_buffer(#buf{} = Buf) ->
 -spec reply_buffered(req()) -> ok.
 reply_buffered(#req{call_ref = CallRef}) ->
   Reply = #brod_produce_reply{ call_ref = CallRef
-                             , offset   = ?BROD_PRODUCE_UNKNOWN_OFFSET
+                             , base_offset = ?BROD_PRODUCE_UNKNOWN_OFFSET
                              , result   = brod_produce_req_buffered
                              },
   cast(CallRef#brod_call_ref.caller, Reply).
@@ -395,7 +395,7 @@ reply_acked(Req) ->
 -spec reply_acked(req(), offset()) -> ok.
 reply_acked(#req{call_ref = CallRef}, Offset) ->
   Reply = #brod_produce_reply{ call_ref = CallRef
-                             , offset   = Offset
+                             , base_offset = Offset
                              , result   = brod_produce_req_acked
                              },
   cast(CallRef#brod_call_ref.caller, Reply).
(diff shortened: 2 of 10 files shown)


aronisstav:
This Travis job needs to be restarted: https://travis-ci.org/klarna/brod/jobs/397265533

zmstone:
The offset in produce response is the base offset of a produced batch.
Messages from multiple `brod:produce` calls may get batched in to one produce request,
hence replying the same offset for all calls may not be optimal.
we can try to calculate offsets for each message.

Please note that a producer may send batches themselves as
```
brod:produce(Client, Topic, Partition, _Key =<<>>, _Value = [{K1, V1}, {K2, V2}]).
% where V1 and V2 can also be nested batches
```
i.e. we'll need to `Offset = LastOffset +1` for each of the inner batches while calculating offsets.

include/brod.hrl
```diff
@@ -44,7 +44,10 @@
                        , ref    :: reference()
                        }).
 
+-define(BROD_PRODUCE_UNKNOWN_OFFSET, -1).
+
 -record(brod_produce_reply, { call_ref :: brod:call_ref()
+                            , offset   :: brod:offset()
```
zmstone: maybe call it `base_offset` as it is in the produce response.

aronisstav: Good point!

src/brod_utils.erl
```diff
@@ -497,6 +498,26 @@ bytes(?TKV(T, K, V)) when is_integer(T) -> 8 + bytes(K) + bytes(V);
...
+
+%% @private Foldl kv-list.
+foldl_kvlist(_Fun, Acc, []) -> Acc;
+foldl_kvlist(Fun, Acc, [Msg | Rest]) ->
+  NewAcc = case nested(Msg) of
+             false -> Fun(Msg, Acc);
+             Nested -> foldl_kvlist(Fun, Acc, Nested)
+           end,
+  foldl_kvlist(Fun, NewAcc, Rest).
+
```
aronisstav, line 520: I copied this code from `kafka_protocol/src/kpro.erl`.

src/brod_producer_buffer.erl
```diff
@@ -137,14 +141,20 @@ maybe_send(#buf{} = Buf, SockPid, Vsn) ->
...
+  ack(Buf, CorrId, ?BROD_PRODUCE_UNKNOWN_OFFSET).
+
+-spec ack(buf(), corr_id(), offset()) ->
+             {ok, buf()} | {error, none | corr_id()}.
 ack(#buf{ onwire_count = OnWireCount
         , onwire       = [{CorrId, Reqs} | Rest]
-        } = Buf, CorrId) ->
-  ok = lists:foreach(fun reply_acked/1, Reqs),
+        } = Buf, CorrId, BaseOffset) ->
+  Fold = fun(R, Offset) -> reply_acked(R, Offset) end,
```
zmstone: `Fold = fun reply_acked/2`

zmstone:
very nice !

src/brod_producer_buffer.erl
aronisstav: True!

viktoriafordos (commented):
~~There are non-backward compatible changes in this PR. 
For instance, sync_produce_request will return {ok, offset()} instead of ok. This will require all client to change their code if they want to upgrade brod. I think this can be avoided. However, I don't know what is the contribution policy in general for the brod library.
Are we fine with being non-backward compatible @zmstone ?~~

@aronisstav pointed out that the API module is brod. The signature of existing functions in brod has not been changed.

Another point is that pending async produces will not be acknowledged even though they were successful because the brod_produce_reply record has changed and is used in a receive. So upgrading the application may yield to false error.

zmstone (approved):

zmstone (changes requested):
Sorry, missed two things:
1. Please bump version to 3.5.0 in both .app.src and Makefile.
2. Please update changelog.md to describe this new feature (and hint that upgrading to this version will require restarts of all producers).

aronisstav:
@zmstone Done requested changes.

Regarding @viktoriafordos comment, the changelog makes the potential issues on upgrade explicit.

zmstone (approved):

test/brod_producer_SUITE.erl
```diff
@@ -34,16 +34,20 @@
...
         , t_producer_topic_not_found/1
         , t_producer_partition_not_found/1
         , t_produce_partitioner/1
         , t_produce_batch/1
+        , t_produce_buffered_offset/1
         ]).
 
 -include_lib("common_test/include/ct.hrl").
 -include_lib("eunit/include/eunit.hrl").
+-include("brod.hrl").
```
aronisstav: I noticed today that this is redundant. Should I remove it?

test/brod_producer_SUITE.erl
```diff
@@ -57,7 +61,7 @@ subscriber_loop(Client, TesterPid) ->
       #kafka_message_set{ messages = Messages
                         , partition = Partition} = KMS,
       lists:foreach(fun(#kafka_message{offset = Offset, key = K, value = V}) ->
-                      TesterPid ! {Partition, K, V},
+                      TesterPid ! {Partition, Offset, K, V},
                       ok = brod:consume_ack(ConsumerPid, Offset)
                     end, Messages),
       subscriber_loop(Client, TesterPid);
```
aronisstav: `Client` is unused in this function (only going to the tail call). Should I remove it?

test/brod_producer_SUITE.erl
zmstone: yes please.

test/brod_producer_SUITE.erl
zmstone: yes please.

zmstone (approved):

k32 (approved):

PR #282, opened 2018-08-14
Allows a topic subscriber to start reading a partition from offset 0

eloraburns:
by allowing the client to pass in an offset of -1. This seems like the
only reasonable way to maintain the old API while allowing the full
breadth of starting offset values.

I hit this case today while testing my application that makes use of compacted topics. On a restart of our `brod_topic_subscriber` process, it successfully re-subscribes to the brod consumer, but didn't re-start consuming from the beginning with `ConsumerConfig = [{begin_offset, earliest}]`. It appears that we have to return `{ok, [{PartitionId, EarliestOffset - 1}], State}` from our `init` to start re-consuming from where we left off. This works nicely…_except_ when the earliest offset is actually 0!

Gross as it is, this shouldn't break dialyzer as kpro:offset seems to be
a signed int.

This seems slightly related to klarna/brod#261, but at the "starting" end of things.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -111,4 +111,7 @@
 * 3.6.1
   * Make produce request version configurable as `produce_req_vsn` in `brod:producer_config()`
   * Upgrade `kafka_protocol` to `2.1.2` to support alpine/busybox build
+* 3.6.2
+  * Allow `brod_topic_subscriber` to explicitly start consuming from partition offset 0 (by passing in
+    a committed offset of -1).
 
--- a/src/brod.app.src
+++ b/src/brod.app.src
@@ -1,7 +1,7 @@
 %% -*- mode:erlang -*-
 {application,brod,
  [{description,"Apache Kafka Erlang client library"},
-  {vsn,"3.6.1"},
+  {vsn,"3.6.2"},
   {registered,[]},
   {applications,[kernel,stdlib,kafka_protocol,supervisor3]},
   {env,[]},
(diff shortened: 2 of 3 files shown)


zmstone:
Hi @taavi 
Thanks for the PR!
An alternative solution is to return `{ok, []}` in the init callback, then
`[{begin_offset, earliest}]` will be used as default.

However, I'm not following why this has anything to do with compacted topic.

zmstone:
Please bump a minor in `brod.app.src` then add a line to `changelog.md`

eloraburns:
@zmstone Unfortunately I've found that returning `{ok, [], State}` from my topic subscriber module does not result in re-reading the topic's full history (which is the intent). Looking at the code, it appears that when the `topic_consumer` re-subscribes to the `brod_consumer`, the previous partition offset is maintained https://github.com/klarna/brod/blob/master/src/brod_consumer.erl#L613-L615, so passing an implicit empty list of offsets doesn't actually reset to the beginning. Empirically, when I pass back an empty list of offsets my consumer does not receive old messages on restart. When I use `resolve_offset` to find the earliest and return it from `init` explicitly, then my consumer does get all of history. Except that it can't get the message at offset 0. ;)

The compacted topic is not interesting from the code perspective, but is why I need the topic subscriber to re-read from the beginning of the topic on restart. :) Though now that I think about it, it would be true of any event-source style topic. So no, it's not relevant.

I'll bump the version and changelog shortly, thanks!

zmstone:
3.6.2 published to hex. thanks! @taavi

eloraburns:
Yay!

PR #283, opened 2018-08-20
New API for asynchronous offset commit

k32:
This new API allows to ack and commit offsets independently.

Rationale: subscribers may want to implement their own buffering logic independent from brod, while still benefiting from brod's group offset management and load balancing. With the current approach it's troublesome: brod consumer message buffer should be made very large, leading to all kinds of problems.

There's a known quirk in the current PR, though: when group_subscriber is started with begin_offset=latest and restarts before committing anything, it will be restarted from the latest offset. Probably it's not the most obvious behavior, though formally correct

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -114,4 +114,6 @@
 * 3.6.2
   * Allow `brod_topic_subscriber` to explicitly start consuming from partition offset 0 (by passing in
     a committed offset of -1).
-
+* 3.7.0
+  * Add `brod_group_subscriber:ack/5` and `brod_group_subscriber:commit/4` to let group subscribers
+    commit offsets asynchronously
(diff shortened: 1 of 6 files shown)


zmstone (commented):
Please bump version in Makefile and brod.app.src (3.7.0)
then add description for this new version in changelog.md.

changelog.md
```diff
@@ -114,4 +114,6 @@
 * 3.6.2
   * Allow `brod_topic_subscriber` to explicitly start consuming from partition offset 0 (by passing in
     a committed offset of -1).
-
+* 3.7.0
+  * Add `brod_group_subscriber:ack/5' and `brod_group_subscriber:commit/4' to let group subscribers
```
zmstone: it's not edoc here, but markdown. quote by `

k32: :see_no_evil:

k32:
Applied the remark and squashed

PR #284, opened 2018-08-20
Handles the case where a message set contains no useful messages.

eloraburns:
This seems related to #275 and deals with additional effects of compacted topics.

In `brod_consumer`'s `handle_batches`, it calls `flatten_batches` which coalesces messages in a list of message sets and then filters out any messages with offests smaller than the originally requested offset. This can result in an empty message set—and if there are no other unacked
offsets for the consumer—a crash.

I've tested this empirically in our test environment (1abff7c72b45b6bb6ed38754d6152eb5c4642374 vs 9bb6717ffc36dd63eb2f20db21c8cad1cefec53c) and it seems to work. :)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -117,3 +117,5 @@
 * 3.7.0
   * Add `brod_group_subscriber:ack/5` and `brod_group_subscriber:commit/4` to let group subscribers
     commit offsets asynchronously
+  * Fix issue #384: In compacted topics, Kafka may return a MessageSet which contains only
+    messages before the desired offset. Just keep reading forward in this case.
--- a/src/brod_consumer.erl
+++ b/src/brod_consumer.erl
@@ -396,21 +396,30 @@ handle_batches(Header, Batches,
                      , partition    = Partition
                      } = State0) ->
   HighWmOffset = kpro:find(high_watermark, Header),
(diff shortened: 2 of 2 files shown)


eloraburns:
Given that I've found a couple of different bugs in this, I wonder if it would be useful to attempt to create an automated test case for things like this. Not sure how easy that would be though; there's a nice return, but a pretty heavy investment and the possibility that it could be pretty flaky, depending on how it's done.

I would probably try to create some pre-baked Kafka data files that can be dumped directly into the broker's data directory in order to ensure that we get the compacted topic layout we expect, but it might be annoying to ensure it works across different broker versions. Any thoughts on this?

Thanks!

zmstone:
@taavi Thanks for the PR.
It should not deliver an empty message set to subscriber though.
perhaps lie this:
```
  case brod_utils:flatten_batches(BeginOffset, Batches) of
    [] ->
      State1 = State0#state{begin_offset = BeginOffset + 1},
      State = maybe_send_fetch_request(State1),
      {noreply, State};
    Messages ->
      ... as usual ...
  end.
```

eloraburns:
I tested these new changes, and it looks like things still work. :)

PR #292, opened 2018-09-25
Allow brod_group_subscriber:assign_partitions/3 to update #state.cb_s…

onno-vos-dev:
Allow brod_group_subscriber:assign_partitions/3 to update #state.cb_state. 

The reason for this being that this callback can be used to handle Leader election switches between different nodes. There may be a scenario where my cb_state has to be cleared. Implementation is backwards compatible :+1:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -125,3 +125,4 @@
   * Fix brod_topic_subscriber and brod_group_subscriber re-subscribe behaviour
     to avoid fetching duplicated messages.
   * Add 'random' and 'hash' partitioner for produce APIs
+  * Allow `brod_group_subscrber:assign_partitions/3` to return an updated cb_state
--- a/src/brod_group_subscriber.erl
+++ b/src/brod_group_subscriber.erl
@@ -379,8 +379,14 @@ handle_call({assign_partitions, Members, TopicPartitions}, _From,
             #state{ cb_module = CbModule
                   , cb_state  = CbState
                   } = State) ->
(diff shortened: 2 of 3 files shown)


changelog.md
```diff
@@ -125,3 +125,5 @@
   * Fix brod_topic_subscriber and brod_group_subscriber re-subscribe behaviour
     to avoid fetching duplicated messages.
   * Add 'random' and 'hash' partitioner for produce APIs
+* 3.7.2
```
zmstone: 3.7.1 is not release yet, so we can release this as a part of 3.7.1

src/brod_group_subscriber.erl
```diff
@@ -379,8 +379,12 @@ handle_call({assign_partitions, Members, TopicPartitions}, _From,
             #state{ cb_module = CbModule
                   , cb_state  = CbState
                   } = State) ->
-  Result = CbModule:assign_partitions(Members, TopicPartitions, CbState),
-  {reply, Result, State};
+  case CbModule:assign_partitions(Members, TopicPartitions, CbState) of
+    {NewCbState, Result} ->
+      {reply, Result, State#state{ cb_state = NewCbState }};
+    Result ->
```
zmstone: please add a `when is_list(Result) ` guar for this clause.
aso add a comment line to explain why we should keep this clause.

onno-vos-dev:
Test can be simplified

zmstone:
💯

onno-vos-dev:
Beautiful! Thank you zmstone!

PR #297, opened 2018-10-26
Use 'begin_offset' from consumer config in brod_group_subscriber

kubaodias:
When brod_group_subscriber is started with `ConsumerConfig = [{begin_offset, N}]` then this configuration option has no effect if any offset is committed to Kafka topic. In such case consumer will always be started from the last committed offset.
If consumer provides `{begin_offset, N}` configuration option then this value should be used as beginning offset.

### DIFF
--- a/src/brod_group_subscriber.erl
+++ b/src/brod_group_subscriber.erl
@@ -576,12 +576,13 @@ do_commit_ack(Pid, GenerationId, Topic, Partition, Offset) ->
 
 subscribe_partitions(#state{ client    = Client
                            , consumers = Consumers0
+                           , consumer_config = ConsumerConfig
                            } = State) ->
   Consumers =
-    lists:map(fun(C) -> subscribe_partition(Client, C) end, Consumers0),
+    lists:map(fun(C) -> subscribe_partition(Client, C, ConsumerConfig) end, Consumers0),
   {ok, State#state{consumers = Consumers}}.
 
-subscribe_partition(Client, Consumer) ->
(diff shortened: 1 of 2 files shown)


zmstone:
Hi @kubaodias 
That config is only intended to work as a default (bootstrap) value when there is no commit history.
We can not make use of this config for each start/restart because
1. The config will be used for each and every group member start/restart, meaning if one would want to use this config once, and after that it should use commit history instead, they would have to set this config, then delete it after it successfully committed new history to kafka.
2. The config is applied for ALL partitions, when there is a need to discard commit history, it's likely that each partition will have different starting offset (i.e. not always `earliest` or `latest`)

There are 2 easier ways to reset commit history:
1. If it's to make all members (all topic-partitions) start from either `earliest` or `latest`, simply switch to a new consumer group ID.
2. If it's to reset commit history with per-topic-partition offset, brod-cli can help (see brod-cli examples in README).

zmstone:
you can provide committed offset in return value of the init callback for `brod_topic_subscriber`

for `brod_group_subscriber` you can set `offset_commit_policy` to `consumer_managed` and implement the `get_committed_offsets` callback.

zmstone:
@kubaodias Sorry the `brod_topic_subscriber` issue may due to a bug introduced in 3.7.1
#299 should fix it.

zmstone:
Then I believe it's consumer group to use.
There is no easy way to reset offset, I do not think putting the offsets to config is a good option because start/restart will make it start from config again.
Workaround is to use a new consumer group or force reset offsets using brod-cli.

PR #298, opened 2018-10-29
Causes a connection reset when unknown_topic_or_partition error happens.

eloraburns:
According to the Kafka Protocol documentation, this is a retriable error:
https://kafka.apache.org/protocol#protocol_error_codes
Error: UNKNOWN_TOPIC_OR_PARTITION
Code: 3
Retriable: True
Description: This server does not host this topic-partition.

This is unlike invalid_topic_exception which is not retriable.

We have concretely experienced this causing our topic consumer to hang because the subscriber quietly went away. The subscriber can be made to reliably go away by rebalancing partitions, such that the broker which was the leader is no longer hosting a replica of the partition in question.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 KAFKA_VERSION ?= 1.1
 PROJECT = brod
 PROJECT_DESCRIPTION = Kafka client library in Erlang
-PROJECT_VERSION = 3.7.1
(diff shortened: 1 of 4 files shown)


zmstone:
@taavi Thanks for the PR. please bump version in `brod.app.src` and update `changelog.md`

PR #299, opened 2018-10-30
On initial subscribe, the last_offset should be the acked_offset.

eloraburns:
This makes sense, as on startup any explicitly specified offset should be considered completely "seen".

Without this, passing partition offsets to the topic subscriber by returning them from CbModule:init results in no subscription, and no consuming.


zmstone:
I would prefer to have it fixed in `subscribe_parittion`
```
    false when AckedOffset =/= LastOffset ->
      %% The last fetched offset is not yet acked,
      %% do not re-subscribe now to keep it simple and slow.
      %% Otherwise if we subscribe with {begin_offset, LastOffset + 1}
      %% we may exceed pre-fetch window size.
      Consumer;
```
change to
```
    false when AckedOffset =/= LastOffset andalso LastOffset =/= ?undef ->

```

zmstone:
this fix may not be needed for `brod_group_subscriber`,
but please help to make the same change there
because the code is more clear this way.

pleas also update changelog.md, then update .app.src with version `3.7.2` same version for #298

eloraburns:
NOTE: I rebased this on top of #298 to avoid merge conflicts in the changelog. Probably merge #298 first, then merging this PR will effectively add only the changes from itself.

zmstone:
@taavi Thanks! Squash merged this one.

zmstone:
`3.7.2` tagged and published hex.

eloraburns:
Thanks! 😄

PR #303, opened 2019-01-19
Add group member callback to provide user_data in group join requests

zmstone:
This is to allow group members to share states with group leader.
Two use case examples:
1. A member may choose to ignore the assignments_revoked callback, and include its currently assigned partitions in user_data, so the leader can apply a 'sticky' assignment hence it may continue working on the current assignments without interruption
2. A member may include its system capacity (nr. of cores, free memory etc), so the leader can assign partitions in a more balanced fashion.
This however all requires users to implement their own assignment strategy.

### DIFF
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -1015,7 +1015,8 @@ make_offset_commit_metadata() ->
 %% if assign_partitions callback is not implemented, otherwise
 %% should be set in the assign_partitions callback implemenation.
 user_data(Module, Pid) ->
-  case lists:member({user_data, 1}, Module:module_info(exports)) of
+  %% Module is ensured to be loaded already
+  case erlang:function_exported(Module, user_data, 1) of
     true -> Module:user_data(Pid);
     false -> <<>>
   end.
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -512,11 +512,13 @@ join_group(#state{ groupId                 = GroupId
                  , connection              = Connection
                  , session_timeout_seconds = SessionTimeoutSec
                  , protocol_name           = ProtocolName
+                 , member_module           = MemberModule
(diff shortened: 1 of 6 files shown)


k32 (approved):

src/brod_group_coordinator.erl
```diff
@@ -1007,12 +1009,16 @@ make_offset_commit_metadata() ->
...
-%% `brod_received_assignments()'
--spec user_data(join | assign) -> binary().
-user_data(_Action) -> <<"">>.
+%% Currently user_data in join requests can be created from a
+%% callback to member module.
+%% user_data assigned back to members is always an empty binary
+%% if assign_partitions callback is not implemented, otherwise
+%% should be set in the assign_partitions callback implemenation.
+user_data(Module, Pid) ->
+  case lists:member({user_data, 1}, Module:module_info(exports)) of
```
mikpe: How frequently is this called?  We used to have noticeable performance issues due to similar code.  Now we prefer to use something like

```
is_exported(M, F, A) ->
    %% make a cheap test first to avoid unnecessary calls to code server
    case erlang:module_loaded(M) of
        false -> code:ensure_loaded(M);
        true -> ok
    end,
    erlang:function_exported(M, F, A).
```

zmstone: Thank you! Now I remembered.
I had a bad feeling about this function but couldn't figure out why when writing it.

Nonetheless, this is not frequent at all.
Also, this becomes a high cost only if the module does not exist?
In our case, the module is ensured to be loaded already, otherwise it can never get this far.

mikpe: If you know the module is loaded, I'd suggest to simply use erlang:function_exported/3.  It's cleaner than inspecting the exports list, and allocates less garbage.

k32 (approved):

mikpe (approved):

PR #341, opened 2019-07-26
Fix handle_message callback spec of group subscriber

chulkilee:
Seems like `brod_group_subscriber` callback `handle_message` may be called with message_set, not just message.

https://github.com/klarna/brod/blob/f45adf912764cd75757783f6881c7a5371d0a001/src/brod_group_subscriber.erl#L505-L522


k32:
Hi! This behavior is intentional, it enables batch processing.

k32 (approved):

k32:
Thanks for the PR!

PR #348, opened 2019-10-10
Create and Delete Topic Apis

v-kat:
I needed create and delete topics for a project I'm working on, so I added them in.
I've tested them from elixir with the following:

```elixir
topic_config = [
  %{
    config_entries: [],
    num_partitions: 2,
    replica_assignment: [],
    replication_factor: 1,
    topic: "test_create_topic"
  }
]
:brod.create_topics([{'localhost', 9092}], topic_config, %{})
:brod.delete_topics([{'localhost', 9092}], ["test_create_topic"], 1_000)
```

and they both worked. Let me know if you want to me to try to add tests or if there's any other issues. Apologies for the readme space removal (hopefully that's not an issue). 

I'm mostly delegating to your kafka_protocol library which does all the work. I wasn't sure about the API version support so I went with `{0, 0}` but am willing to change it if that's incorrect.

Thanks for making brod!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ script:
       # run dialyzer only in otp 22
       rebar3 dialyzer
       # test against kafka 0.9 in otp 22
-      export KAFKA_VERSION="0.9"
+      export KAFKA_VERSION="1.0"
     fi
     if [ $OTP_VSN -eq 21 ]; then
       # test against kafka 2.2 in otp 21
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ script:
       # run dialyzer only in otp 22
       rebar3 dialyzer
       # test against kafka 0.9 in otp 22
-      export KAFKA_VERSION="1.0"
+      export KAFKA_VERSION="0.9"
     fi
     if [ $OTP_VSN -eq 21 ]; then
       # test against kafka 2.2 in otp 21
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -78,6 +78,11 @@
         , start_link_topic_subscriber/7
         ]).
 
+%% Topic APIs
+-export([ delete_topics/3
+        , delete_topics/4
+        ]).
+
 %% APIs for quick metadata or message inspection and brod_cli
 -export([ get_metadata/1
         , get_metadata/2
@@ -739,6 +744,22 @@ start_link_topic_subscriber(Client, Topic, Partitions,
   brod_topic_subscriber:start_link(Client, Topic, Partitions,
                                    ConsumerConfig, MessageType,
                                    CbModule, CbInitArg).
+
+%% @doc delete topic(s) from kafka
+%% Return the message body of `delete_topics, response.
+%% See `kpro_schema.erl' for struct details
+-spec delete_topics([endpoint()], [topic()], pos_integer()) ->
+        {ok, kpro:struct()} | {error, any()}.
+delete_topics(Hosts, Topics, Timeout) ->
+  brod_utils:delete_topics(Hosts, Topics, Timeout).
+
+%% @doc delete topic(s) from kafka
+%% Return the message body of `delete_topics, response.
+%% See `kpro_schema.erl' for struct details
+-spec delete_topics([endpoint()], [topic()], pos_integer(), conn_config()) ->
+        {ok, kpro:struct()} | {error, any()}.
+delete_topics(Hosts, Topics, Timeout, Options) ->
+  brod_utils:delete_topics(Hosts, Topics, Timeout, Options).
 
 %% @doc Fetch broker metadata
 %% Return the message body of `metadata' response.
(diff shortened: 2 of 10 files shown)


src/brod.erl
```diff
@@ -740,6 +748,40 @@ start_link_topic_subscriber(Client, Topic, Partitions,
                                    ConsumerConfig, MessageType,
                                    CbModule, CbInitArg).
 
+%% @doc create topic(s) in kafka
+%% Return the message body of `create_topics, response.
+%% See `kpro_schema.erl' for struct details
```
k32: Nitpick: Consider using `@equiv` edoc tag [1] to avoid duplicating the docstring. Same applies to the other new functions in this module

[1] http://erlang.org/doc/apps/edoc/chapter.html#function-tags

k32:
Hi!

Something happened to our webhooks, and Travis didn't update the PR. I am looking into this. But newly added testcases failed: https://travis-ci.org/klarna/brod/jobs/596728943 Could you please take a look?

test/brod_SUITE.erl
```diff
@@ -0,0 +1,82 @@
...
+        ]).
+
+%% Test cases
+-export([ t_create_topics/1
+        , t_delete_topics/1
+        , t_delete_topics_not_found/1
+        ]).
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("eunit/include/eunit.hrl").
```
k32: If this is just for asserts then include `"stdlib/include/assert.hrl"` instead

src/brod.erl
```diff
@@ -740,6 +748,36 @@ start_link_topic_subscriber(Client, Topic, Partitions,
...
 
+%% @equiv create_topics(Hosts, TopicsConfigs, RequestConfigs, [])
+-spec create_topics([endpoint()], [topic_config()], #{timeout => kpro:int32(),
+                    validate_only => boolean()}) ->
+        {ok, kpro:struct()} | {error, any()}.
+create_topics(Hosts, TopicConfigs, RequestConfigs) ->
+  brod_utils:create_topics(Hosts, TopicConfigs, RequestConfigs).
+
+%% @doc create topic(s) in kafka
+%% Return the message body of `create_topics, response.
```
k32: Missing `'` after create_topics?

v-kat:
@k32 I addressed your comments and also fixed the elvis lint errors. I removed the suite from the `make it` command as it kept timing out when trying to connect to the kafka controller node on the CI for some reason https://travis-ci.org/klarna/brod/jobs/596803914 shows it passing though 😿

test/brod_SUITE.erl
```diff
@@ -0,0 +1,88 @@
...
+%% Test cases
+-export([ t_create_topics/1
+        , t_delete_topics/1
+        , t_delete_topics_not_found/1
+        ]).
+
+-include_lib("common_test/include/ct.hrl").
+-include_lib("stdlib/include/assert.hrl").
+
+-define(HOSTS, [{"localhost", 9092}, {"localhost", 9192}]).
```
k32: Remove the second host, I believe it's the reason behind tiimeouts in CI.

k32:
Hi @IRog ! We have a requirement that all API functions should be tested in CI, so this timeout issue needs to be addressed. But I am positive that this happens because Kafka listens on different ports (plain text: 9092, ssl: 9093, ssl + sasl: 9094) in our CI environment. So it should be an easy fix.

v-kat:
@k32 the timeout is still happening even with the config change. I have no idea what causes the issue but everything works locally and it sometimes passes on travis.

test/brod_SUITE.erl
v-kat: done but the issue is still occurring

v-kat:
@zmstone could you please build the docker image for `0.9`? https://hub.docker.com/r/zmstone/kafka/tags shows that it hasn't been updated and now all the tests are passing minus https://travis-ci.org/klarna/brod/jobs/597790256 which uses that image. Also it's kind of sad that the timeout error was returned from not having an option enabled. I think that's at the kafka protocol level though :(. Sorry for all the commits but hopefully you can `squash and merge` when putting this in.

zmstone:
I don't think 0.9 (maybe 0.10 also) supports topic create/delete APIs

.travis.yml
```diff
@@ -43,7 +43,7 @@ script:
       # run dialyzer only in otp 22
       rebar3 dialyzer
       # test against kafka 0.9 in otp 22
-      export KAFKA_VERSION="0.9"
+      export KAFKA_VERSION="1.0"
```
zmstone: I was not clear enough.
kafka 0.9 docker image is no longer updated
but that does not imply we do not support 0.9 any more in `brod`

we can instead check kafka version (by e.g. query api version) in test case,
and simply *pass* it if it's not 1.0 or later.

k32 (approved):

k32:
Thanks! It looks good to me. Will keep the ticket open for a day in case zmstone or someone else has something to add. Edit: or not

zmstone (approved):

PR #351, opened 2019-10-24
Receive pending acks after assignments_revoked is invoked

josevalim:
We are building a multi-process data processing pipeline
with Brod and we are using the assignments_revoked callback
to flush all messages being processed. However, if any message
is acknowledged during the assignments_revoked step, those
acks won't be committed, they will just sit in the group coordinator
inbox and eventually be processed too late.

This commit makes it so we retrieve any pending ack from the
inbox before we commit them.

### DIFF
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -287,14 +287,7 @@ init({Client, GroupId, Topics, Config, CbModule, MemberPid}) ->
   {ok, State}.
 
 handle_info({ack, GenerationId, Topic, Partition, Offset}, State) ->
-  case GenerationId < State#state.generationId of
-    true  ->
-      %% Ignore stale acks
-      {noreply, State};
-    false ->
-      {ok, NewState} = handle_ack(State, Topic, Partition, Offset),
-      {noreply, NewState}
-  end;
+  {noreply, handle_ack(State, GenerationId, Topic, Partition, Offset)};
 handle_info(?LO_CMD_COMMIT_OFFSETS, #state{is_in_group = true} = State) ->
   {ok, NewState} =
     try
@@ -434,29 +427,36 @@ stabilize(#state{ rejoin_delay_seconds = RejoinDelaySeconds
   %% 1. unsubscribe all currently assigned partitions
   ok = MemberModule:assignments_revoked(MemberPid),
 
-  %% 2. try to commit current offsets before re-joinning the group.
+  %% 2. some brod_group_member implementations may wait for messages
+  %%    to finish processing when assignments_revoked is called.
+  %%    The acknowledments of those messages would then be sitting
+  %%    in our inbox. So we do an explicit pass to collect all pending
+  %%    acks so they are included in the best-effort commit below.
+  State1 = receive_pending_acks(State0),
+
+  %% 3. try to commit current offsets before re-joinning the group.
   %%    try only on the first re-join attempt
   %%    do not try if it was illegal generation exception received
   %%    because it will fail on the same exception again
-  State1 =
+  State2 =
     case AttemptNo =:= 0 andalso
          Reason    =/= ?illegal_generation of
       true ->
-        {ok, #state{} = State1_} = try_commit_offsets(State0),
-        State1_;
+        {ok, #state{} = State2_} = try_commit_offsets(State1),
+        State2_;
       false ->
-        State0
+        State1
     end,
-  State2 = State1#state{is_in_group = false},
+  State3 = State2#state{is_in_group = false},
 
-  %$ 3. Clean up state based on the last failure reason
-  State = maybe_reset_member_id(State2, Reason),
+  %$ 4. Clean up state based on the last failure reason
+  State = maybe_reset_member_id(State3, Reason),
 
-  %% 4. ensure we have a connection to the (maybe new) group coordinator
+  %% 5. ensure we have a connection to the (maybe new) group coordinator
   F1 = fun discover_coordinator/1,
-  %% 5. join group
+  %% 6. join group
   F2 = fun join_group/1,
-  %% 6. sync assignemnts
+  %% 7. sync assignemnts
   F3 = fun sync_group/1,
 
   RetryFun =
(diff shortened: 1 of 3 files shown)


k32:
Hi! Thanks for your PR, I need to take a closer look. Could you please add a test case for this scenario in the meanwhile?

josevalim:
Hi @k32! Indeed I was planning to look at the tests today. This bug only triggers during rebalancing (assignments revoked), so do you currently have any tests in the suite that use rebalancing that I can use as a starting point? Thanks.

k32:
Good question. I think the closest one would be `brod_group_subscriber_SUITE:t_2_members_one_partition()` This testcase starts two group subscribers competing for a partition, although there's no re-balancing and no custom-callbacks.

P.S. I started refactoring the testcases using https://github.com/klarna/snabbkaffe library, you can find informal description here: https://k32.github.io/posts/2019-10-11-hello-world.html , if it helps.

josevalim:
Hi @k32, I have pushed a test.

I am not using the `brod_group_subscriber` but I took a cursory glance there and it may be that you have an ack-related bug there too, which can cause duplicated messages. When `assignments_revoked` in `brod_group_subscriber` is called, it seems the consumers are simply unsubscribed without waiting for the conclusion of any message that they may be processing. So by the time they fully process the message, the assignments were revoked, and the acks are no longer useful.

I may be wrong, it was just a cursory glance, but it may be worth a look. In any case, fixing the `brod_group_coordinator`is a necessary step for fixing the subscribers, so I would appreciate if this PR can be verified and merged. Thank you!

zmstone:
Hi @josevalim Nice fix!
However you may have noticed that the commit before
group rebalancing is a best effort, fire-n-forget attempt.
My little concern here is: it may not really fix your issue.

If this has been proven working well,
I'd like to have some code comments describing the case.
e.g. which kafka version,  what was the cause of the rebalance ?
was it an error code in heartbeat response? which error code ?

zmstone:
Some comments with regards to `brod_group_subscriber`.

'at least once delivery' has lead to few relaxed
design with regards to ack/commit handling.
If we are to make `brod_group_subscriber` wait for async acks
before replying the revoke call, we should wait with a timeout.
Long message processing callbacks has been a trouble maker
for group stability (especially in the early version java library,
they had to run a heartbeat thread for each consumer group).

In some cases when group rebalance timeout, members run in
join-backoff-rejoin loop at a certain pace, the group may not be
able to reach balance for very long time.

josevalim:
Hi @zmstone!

Yes, I am aware the commit when assignments are revoked is a fire and forget. The intention of this patch is only to make the "best-effort" a bit better. :)

> If this has been proven working well, I'd like to have some code comments describing the case.

Good call. I can definitely add a bit more context. The cause of the rebalancing was simply starting new Kafka consumers. Our group implementation does perform flushing/draining when assignments are revoked, which is why this patch is necessary. This is the scenario we emulate in the added test and we have verified the patch works both in tests and in practice.

I can also include code comments about the Kafka version but I am not sure it would be a good idea. Adding the Kafka version may imply we are addressing an issue that is specific to the Kafka server but this is a client specific concern.

> Some comments with regards to brod_group_subscriber. 'at least once delivery' has lead to few relaxed design with regards to ack/commit handling.

Thank you for the context. This clarifies a lot!

test/brod_group_coordinator_SUITE.erl
```diff
@@ -0,0 +1,127 @@
...
+
+-include_lib("snabbkaffe/include/ct_boilerplate.hrl").
+-include("brod.hrl").
+
+%%%_* ct callbacks =============================================================
+
+suite() -> [{timetrap, {seconds, 60}}].
+
+init_per_suite(Config) ->
+  {ok, _} = application:ensure_all_started(brod),
```
k32: Better move it to `common_init_per_testcase` (and stop brod app in `common_end_per_testcase`). Brod has tendency to start processes under its own supervisor, which may lead to lingering children. Even though this is not relevant in this testsuite yet, it's always good to keep things clean from the very beginning

test/brod_group_coordinator_SUITE.erl
```diff
@@ -0,0 +1,127 @@
...
+%%%   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+%%%   See the License for the specific language governing permissions and
+%%%   limitations under the License.
+%%%
+
+%% @private
+-module(brod_group_coordinator_SUITE).
+-define(CLIENT_ID, ?MODULE).
+-define(OTHER_CLIENT_ID, other_coordinator_id).
+-define(TOPIC, <<"brod-group-subscriber-1">>).
```
k32: This topic name collides with the `group_subscriber_SUITE`, please rename to something unique

k32:
Hi again! CI is failing, could you please take a look: https://travis-ci.org/klarna/brod/builds/603295920

src/brod_group_coordinator.erl
```diff
@@ -579,13 +588,15 @@ sync_group(#state{ groupId       = GroupId
       [format_assignments(TopicAssignments)]),
   start_offset_commit_timer(NewState).
 
--spec handle_ack(state(), brod:topic(), brod:partition(), brod:offset()) ->
-        {ok, state()}.
-handle_ack(#state{ acked_offsets = AckedOffsets
-                 } = State, Topic, Partition, Offset) ->
+-spec handle_ack(state(), brod:group_generation_id(), brod:topic(), brod:partition(), brod:offset()) ->
```
k32: Very long line. We have (unfortunately unwritten) 80 char rule, that will break our internal CI. I will create an Elvis check some day

k32 (changes requested):

k32:
There are some minor remarks, otherwise the PR looks reasonable to me.

src/brod_group_coordinator.erl
josevalim: Elvis actually did check it, that's why the CI failed. :D

It may be worth though adding an elvis command to make, so it is easier to get this feedback locally!

josevalim:
Thank you for the review. I believe I have addressed all feedback!

k32 (approved):

k32:
Thanks for your PR and investigation!

josevalim:
Thanks for merging! One last question, so we can plan our release schedule accordingly, is there any ETA for publishing a new brod release? We also plan to use the recently added APIs for managing topics, so having that out would be helpful too!

k32:
> is there any ETA for publishing a new brod release

There is no such thing. It more or less boils down to either gathering a critical mass of features or following a request from some of the users. If you need some feature released sooner, just let me know.

josevalim:
I see, thanks! So we would definitely appreciate a new release with at least this bug fix, so we can publish our package to Hex, but meanwhile we will gladly use the git version!

k32:
3.9.1 is released.

PR #353, opened 2019-11-01
Make brod work with erlang.mk + hex

srijan:
Enables using as a dependency with erlang.mk via hex.pm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ EDOC_OPTS = preprocess, {macros, [{build_brod_cli, true}]}
 
 ## Make app the default target
 ## To avoid building a release when brod is used as a erlang.mk project's dependency
-app:: vsn-check
+app::
 
 include erlang.mk
 
--- a/changelog.md
+++ b/changelog.md
@@ -184,3 +184,6 @@
   * Added API for topic deletion and creation
 * 3.9.1
   * Receive pending acks after assignments_revoked is invoked
+* 3.9.2
+  * Remove vsn-check dependency from default Makefile target
+    This enables using brod with erlang.mk + hex
--- a/changelog.md
+++ b/changelog.md
@@ -185,5 +185,7 @@
 * 3.9.1
   * Receive pending acks after assignments_revoked is invoked
 * 3.9.2
+  * Fix corrupted package published to hex
+* 3.9.3
   * Remove vsn-check dependency from default Makefile target
     This enables using brod with erlang.mk + hex


k32:
Hi! Thanks for your PR. Do you use brod with Erlang.mk? I am wondering because actually we considered removing support for this build system in the future.

srijan:
> Hi! Thanks for your PR. Do you use brod with Erlang.mk? I am wondering because actually we considered removing support for this build system in the future.

We use erlang.mk, and have started using brod recently.

I think you can safely remove support for building with erlang.mk and shift to rebar3 without affecting our use case.

Erlang.mk supports using rebar3 compatible libraries as dependencies.

k32:
Do you need some particular script to build brod with Erlang.mk? I realized that some of the scripts in the "scripts" directory should not be exposed to the outer world: they are for CI, testing and what not.

srijan:
@k32, the Makefile target for `compile` has `vsn-check` as dependency:

```
vsn-check:
	@./scripts/vsn-check.sh $(PROJECT_VERSION)
```

I didn't go too deep into what it does. If it's not needed, another way to resolve this would be to remove this dependency.

k32:
Thanks for the answer. I think vsn_check dependency can be simply removed from the Makefile. Recently I made some  changes that make rebar3 use git tag as a source of truth. So significance of this script has diminished.

srijan:
It looks like the `vsn-check.sh` script just compares versions of dependencies defined in Makefile and in rebar.config, and exits with an error if there is a mismatch.

If you're only using rebar for building, then I think this script can be removed completely.

k32:
Agree.

srijan:
Should I amend this PR to completely remove erlang.mk (while still keeping it compatible to be used as a dependency in an erlang.mk project) ?

Or should I just make changes required to make it work for now as an erlang.mk project dependency by removing vsn check script?

k32:
I think it's better to start by removing version check dependency in the makefile first. Removing support for Erlang.mk alltogether is a breaking change that deserves a major version bump.

src/brod.app.src
```diff
@@ -11,5 +11,5 @@
   {licenses, ["Apache License 2.0"]},
   {links, [{"Github", "https://github.com/klarna/brod"}]},
   {build_tools, ["make", "rebar", "rebar3"]},
-  {files, ["src", "include", "rebar.config", "rebar.config.script",
+  {files, ["src", "include", "scripts", "rebar.config", "rebar.config.script",
```
k32: Internal scripts should not be packaged

k32 (changes requested):

srijan:
There's still an issue.

The default target in the Makefile is `app`, which uses erlang.mk to build.
In the current state:

1. If a repo using erlang.mk includes brod as a dependency using git -> it builds using erlang.mk (using dependency details specified in Makefile)
2. If a repo using erlang.mk includes brod as a dependency using hex -> build fails because `vsn-check.sh` does not exist in hex package (this is where I got stuck)
3. If a repo using rebar3 includes brod as a dependency -> it builds using rebar3 (using dependency details specified in rebar.config)

The `vsn-check.sh` script ensures that the dependency versions (for supervisor3 and kafka_protocol) defined in rebar.config and Makefile do not go out of sync.

_So, as long as the dependency version is defined both in the Makefile and rebar.config, `vsn-check.sh` should not be removed._

**Solution 1:** include the `vsn-check.sh` file in the hex package. (Not recommended)

**Solution 2:** Only define dependencies in rebar.config. Erlang.mk supports reading dependency information from rebar.config. But, for this, erlang.mk would have to be completely removed from this repo. If erlang.mk file is present, it assumes dependencies are defined in Makefile.

**Solution 3:** (recommended)
Remove vsn-check from the Makefile dependency of `app` and `compile` targets, while still keeping it in the `t` target (used for testing, called in travis.yml).
This way, all 3 cases above will work, and travis will still complain if Makefile and rebar.config dependencies ever go out of sync.

I recommend solution 3 for the short term, and solution 2 can be planned for the next major version bump.

@k32, please confirm, and I'll update the diff accordingly.

k32:
Hi!
You are correct. Option 3 is the best short-term solution.

zmstone:
vsn-check is already a dependency for target `t`
Suggestion: change 
`app:: vsn-check`
To
`app::`
Otherwise 
`app:: dummy`

k32 (approved):

k32:
Hi! Looks good to me. Please update the changelog.

k32 (approved):

k32:
Thanks for the PR!

k32:
@srijan Brod 3.9.3 that includes your patch has been released

PR #362, opened 2019-11-28
Handle undefined fetcher in fold loop exception

zmstone:
Prior to this commit, fold loop may crash if there
is an exception in message handler callback when
processing the last batch (i.e. when there is no
running fetcher for the next batch), kill_fetcher
function may crash on 'undefined' clause


k32 (approved):

mikpe (approved):

zmstone:
merge ?

mikpe:
Merged and tagged as 3.9.4.

PR #365, opened 2020-01-17
Delete unused variables

zmstone:
otherwise does not build in otp 22


qzhuyan:
Isn't it the same as in https://github.com/klarna/brod/commit/33f8044a2634a6e1962980e54ab393b618964867 ?

PR #371, opened 2020-03-13
topic_subscriber and group_subscriber should own their consumers

k32:
Problem: brod looks up subscriber using topic and partition as a key. Any attemt to subscribe to the same topic-partition in the same client will result in error


zmstone:
The limitation is due to `brod_consumer` needs to keep a fetch-ahead buffer, it also keeps track of the acked offsets etc, if to allow multiple subscribers, its internal state has to be reimplemented, and the subscribe API will need to accept an identifier (instead of just a pid which is volatile)
One idea was to spawn a `brod_consumer` for a newer version `brod_group_subscriber` instead of managing the subscriber under `bord_consumer_sup`.
I assumed that the `_v2` subscriber was implemented this way.

k32:
@zmstone I am currently trying out a quick&dirty solution, where consumer key is extended with a "tag", that is an arbitrary term. I.e. instead of looking up consumers by `{topic, partition}`, one can look them up by `{group_id, topic, partition}` instead. In theory, every group_consumer will get its own consumers and they won't compete.

P.S. The old group_subscriber is also vulnerable to this issue, and I don't want to change anything about it on the fundamental level. Hence I want to try this approach first

zmstone:
Sure you can do that.
However, given that the consumer identifier is passed down the supervision tree, and the change have to make it backward compatible etc...
I might be wrong, but I think it's easier to spawn `brod_consumer` and link it to `bord_group_subscriber_v2`.
`brod_consumer` is already extended to work standalone (i.e. not managed by `brod_consumers_sup`, and not register to client) if the `Bootstrap` arg to `start_link/4,5` is set to `undefined`.

k32:
group_subscriber_v2 is actually based on topic_subscriber behavior, that's where it inherit the problem from. (The idea is to spawn and stop topic_subscriber processes on comand from group_coordinator instead of rolling out another implementation of consumer behavior). So the change that you propose should be applied to topic_subscriber. I will check if something can be changed about this. BTW, what was the reason behind having a supervisor tree for the consumers? Does it add something to fault-tolerance?

As for my solution, I think this can be backward compatible on the API level, although live patching will be practically impossible due to changes in supervision trees and internal tables, as you said. Restarting brod during upgrade will work. Not great, not terrible

P.S. Thanks for feedback!

PR #373, opened 2020-03-23
Improve testbed

k32:
Splitting my other PR into manageable chunks. Since I am planning to change how group_subscriber spawn consumers, relying on mocking `brod:start_consumer` is no longer viable.

* Create topics dynamically
* Clean topics before running the tests
* Use a separate client for producing test data
* Start group_consumers from begin_offset=0 instead of latest offset
* Reduce the amount of copy-pasted code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ _rel/
 *.log
 relx
 docker/
+TAGS
--- a/rebar.config
+++ b/rebar.config
@@ -14,4 +14,4 @@
 {cover_enabled, true}.
 {cover_opts, [verbose]}.
 {cover_export_enabled, true}.
-{ct_readable, false}.
+%{ct_readable, false}.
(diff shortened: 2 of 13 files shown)


k32:
@zmstone Could you please take a look, if you have time? No functional change in this PR

rebar.config
```diff
@@ -14,3 +14,4 @@
 {cover_enabled, true}.
 {cover_opts, [verbose]}.
 {cover_export_enabled, true}.
+%{ct_readable, false}.
```
mikpe, line 17: Why add commented-out stuff?

.gitignore
```diff
@@ -23,3 +23,4 @@ _rel/
 *.log
 relx
 docker/
+TAGS
```
mikpe, line 26: nit: missing newline

mikpe (approved):

rebar.config
k32: I found myself turning this option on and off pretty frequently while troubleshooting tests. Maybe it will save someone a few keystrokes too

.gitignore
k32: I will fix it the next PR

Issue #381, opened 2020-04-02
Potential problems during live upgrade because of using supervisor3


tothlac:
Our releases fully support live upgrades between different versions. I've seen this error message when a live upgrade was called:

```
|Mar 23, 2020 @ 16:56:54.559|release_handler: cannot find top supervisor for application brod
```

This error message comes from release_handler_1.erl. release_handler_1:get_supervised_procs/0 function collects data about all supervised modules. Without this error the output of this function would be as follows:

```
2> rp(lists:usort(release_handler_1:get_supervised_procs())).
[{undefined,undefined,<0.48.0>,[kernel]},
 {undefined,undefined,<0.84.0>,[ssl_sup]},
 {undefined,undefined,<0.101.0>,[inets_sup]},
 {undefined,undefined,<0.208.0>,[brod_sup]},
 {<0.48.0>,code_server,<0.49.0>,[code]},
...
```

Unfortunately with the error the supervisor of brod does not appear in this list:

```
2> rp(lists:usort(release_handler_1:get_supervised_procs())).
<0.208.0>: reason: {'EXIT',{{badrecord,state},[{supervisor,get_callback_module,1,[{file,[115,117,112,101,114,118,105,115,111,114,46,101,114,108]},{line,273}]},{release_handler_1,get_supervisor_module,1,[{file,[114,101,108,101,97,115,101,95,104,97,110,100,108,101,114,95,49,46,101,114,108]},{line,684}]},{release_handler_1,get_master_procs,3,[{file,[114,101,108,101,97,115,101,95,104,97,110,100,108,101,114,95,49,46,101,114,108]},{line,605}]},{lists,foldl,3,[{file,[108,105,115,116,115,46,101,114,108]},{line,1263}]},{erl_eval,do_apply,6,[{file,[101,114,108,95,101,118,97,108,46,101,114,108]},{line,684}]},{erl_eval,expr_list,6,[{file,[101,114,108,95,101,118,97,108,46,101,114,108]},{line,888}]},{erl_eval,expr,5,[{file,[101,114,108,95,101,118,97,108,46,101,114,108]},{line,411}]},{erl_eval,expr_list,6,[{file,[101,114,108,95,101,118,97,108,46,101,114,108]},{line,888}]}]}}
2020-04-02T11:04:23.703591+02:00 error: release_handler: cannot find top supervisor for application brod
[{undefined,undefined,<0.48.0>,[kernel]},
[...]

PR #382, opened 2020-04-02
(#381): fix potential live upgrade problem in brod_sup

tothlac:


mikpe (changes requested):
No, this would lead to inconsistencies, the code must either use plain supervisor, or supervisor3, never mix the two.  There are two questions that need answering:
1. Why is brod using supervisor3?  I wasn't involved with brod's initial development, and the git log doesn't really say.  @zmstone @id can either of you shed some light on this?
2. Why doesn't supervisor3 work for the live upgrade case you reported?  That would possibly be a bug in supervisor3, not in brod.

zmstone:
supervisor3 is a copy of supervisor2 from rabbitmq
in fact the repo is a fork of rabbitmq but with only supervisor2 code kept and renamed to supervisor3 to avoid module clashing.
Why we needed supervisor3 was because we like the delayed restart feature.

from the issue description, it seems to me that the tool failed to recognise supervisor3 as a legit supervisor module.

mikpe:
Thanks @zmstone and @robertoaloi, I kind of suspected supervisor3 had gone stale and missed some newer glue with OTP, though I didn't suspect it was _that_ raw and low-level.  I'll take a look at updating supervisor3 unless someone else beats me to it.

robertoaloi:
Hi @mikpe . There is a PR with a patch from @tothlac that fixes the direct issue. See https://github.com/klarna/supervisor3/pull/22

In there, I'm proposing some options for the long-term plans for supervisor3. Please let me know what do you think there.

PR #384, opened 2020-04-06
Update code examples in README.md

bjosv:
I found some small things to tidy up while looking into how to set and modify topic configurations.
I also added the example of how to set one config entry while manually creating the topic, it might help others.
Thanks!


mikpe (approved):

mikpe:
Does this look good to you @zmstone ?

zmstone:
lgtm

PR #385, opened 2020-04-14
Update test to please DRY rule check in elvis

bjosv:
This corrects an issue found during the Travis build.

When Travis is running it uses the latest version of rebar3 and elvis,
but the latest elvis has a change regarding how it handles macros.
Elvis depends on katana_code, which has replaced the pre-processor handling, see:
https://github.com/inaka/katana-code/pull/31

This has changed the behavior of elvis to also include code defined in TEST scope in its checks,
which this PR simply fixes.
This could also be solved by pinning elvis to a specific older version..


mikpe (approved):

bjosv:
Closing this due to its moved into #386 to get all baseline corrections at once.

PR #386, opened 2020-04-15
Corrections to get build checks passed

bjosv:
### Pin version of test dependency snabbkaffe
This avoids breakage due to test framework development,
and will enable the common test suites again.

### Update test to please DRY rule check in elvis 
This corrects an issue found during the Travis build, from #385

When Travis is running it uses the latest version of rebar3 and elvis,
but the latest elvis has a change regarding how it handles macros.
Elvis depends on katana_code, which has replaced the pre-processor handling, see:
inaka/katana-code#31

This has changed the behavior of elvis to also include code defined in TEST scope in its checks,
which this PR simply fixes.
This could also be solved by pinning elvis to a specific older version..

### DIFF
--- a/rebar.config.script
+++ b/rebar.config.script
@@ -2,7 +2,7 @@ IsRebar3 = erlang:function_exported(rebar3, main, 1),
 DocoptUrl = "https://github.com/zmstone/docopt-erl.git",
 DocOptTag = "0.1.3",
 DocoptDep = {docopt, {git, DocoptUrl, {branch, DocOptTag}}},
-Snabbkaffe = {snabbkaffe, {git, "https://github.com/klarna/snabbkaffe.git", {branch, "master"}}}, % TODO: use release tag
+Snabbkaffe = {snabbkaffe, "0.3.3"},
 Profiles =
   {profiles, [
     {brod_cli, [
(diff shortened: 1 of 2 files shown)


k32 (approved):

mikpe (approved):

mikpe:
It would be nice if someone could fix the elvis breakage...

k32:
@mikpe As a repo owner you can force merge this PR, then deal with elvis checks separately

k32:
Thanks for the PR! Alternatively, this dependency can be fetched from hex: https://hex.pm/packages/snabbkaffe

bjosv:
Ah thanks, I didn't want to change the source in case there was a reason for it.
Maybe there can be a commit changing all deps to hex if we want that?
I guess I cant help getting this merged unless the #385 is merged first?..and vice versa

k32:
Hi!

> Ah thanks, I didn't want to change the source in case there was a reason for it.

I don't think there is a good reason to keep this dependency as a source. If I remember correctly, this package hadn't been published to Hex at the time.

> I guess I cant help getting this merged unless the #385 is merged first?..and vice versa

I guess both PRs are needed to get a green build. Order should not be important.

k32 (approved):

bjosv:
> I don't think there is a good reason to keep this dependency as a source. If I remember correctly, this package hadn't been published to Hex at the time.

Sound reasonable and I changed this..much better.

> I guess both PRs are needed to get a green build. Order should not be important.

Since its was two small PRs I merged them into this, seems that I got a coverage run now atleast.
Hope it makes it easier and not only confusing :)

k32:
Thank you!

jesperes (approved):

PR #388, opened 2020-04-18
Do not try to commit offsets after unknown member id error

dszoboszlay:
The group coordinator tries to commit offsets upon receiving an error, except for those errors where we know the commit would fail as well. So far only illegal generation was known to behave like that,
but unknown member id errors are similar too.


k32 (approved):

mikpe (approved):

dszoboszlay:
Interestingly, I get the same CT errors locally, but even for the `master` branch. I have no idea what could be the real reason for that.

zmstone:
rebased, we'll see

PR #392, opened 2020-04-30
Make brod a compatible client with Microsoft's EventHub.

hoarf:
The field "aborted_transaction" seems to contain nil instead of an empty list in the message from MS's EventHub which causes an error in the drop_aborted function.

I don't know if there is any intention of keeping compatibility with EventHub or if this is the most appropriate fix. Let me know if there is anything else I can do.

### DIFF
--- a/src/brod_utils.erl
+++ b/src/brod_utils.erl
@@ -620,6 +620,9 @@ kill_fetcher({Pid, Mref}) ->
       ok
   end.
 
+drop_aborted(#{aborted_transactions := undefined}, Batches) ->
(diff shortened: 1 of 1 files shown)


ntenczar:
Hi! We noticed a similar issue today, though we are not using Microsoft EventHub, and this patch solved our issues. Great find, would love to see this make it into a brod release at some point (a _broader_ release perhaps)

zmstone (approved):

mikpe (approved):

haljin:
We have the exact same problem which is blocking us quite badly. Please let us know when this gets released on Hex :)

PR #393, opened 2020-05-01
Adding resolve_offset/6 to brod_utils, as a way to pass a custom time…

escobera:
…out to kpro:connect_partition_leader


zmstone:
Please also export the API from brod.erl

src/brod_utils.erl
```diff
@@ -141,8 +143,17 @@ get_metadata(Hosts, Topics, ConnCfg) ->
                      offset_time(), conn_config()) ->
         {ok, offset()} | {error, any()}.
 resolve_offset(Hosts, Topic, Partition, Time, ConnCfg) ->
+  Opts = #{timeout => proplists:get_value(
+                                  connect_timeout, ConnCfg, ?DEFAULT_TIMEOUT)},
```
zmstone: elvis is not happy about this line being too long (see travis build),
maybe add a variable for the fallback timeout value.

escobera:
> Please also export the API from brod.erl

Hm, that got me thinking. Will `brod_utils:resolve_offset/6` be public? I don't really know much about erlang so maybe I'm getting this wrong, but this seems to be a private method which only serves as a way to pass connect_timeout property.

In any case I tried exposing it but couldn't get it to pass the dialyzer test. Could you give me some guidance here?

zmstone:
Hi. 
`brod_utils` is a private (by convention) module although there is no way to stop other applications from calling its exported functions directly.
Most of the public APIs in `brod` are exported from brod.erl, the calls to other modules can be simply a delegation.
Worth pointing out that there is no runtime overhead for such delegation thanks to compiler optimisation.

zmstone:
Should be added to the export list in bord.erl.
https://github.com/klarna/brod/blob/master/src/brod.erl#L94

zmstone:
and add a test case to cover the new code:
e.g.
https://github.com/klarna/brod/blob/master/test/brod_consumer_SUITE.erl#L357

src/brod.erl
```diff
@@ -850,7 +850,18 @@ resolve_offset(Hosts, Topic, Partition, Time) ->
                      offset_time(), conn_config()) ->
         {ok, offset()} | {error, any()}.
 resolve_offset(Hosts, Topic, Partition, Time, ConnCfg) ->
-  brod_utils:resolve_offset(Hosts, Topic, Partition, Time, ConnCfg).
+  Timeout =
```
zmstone: this can be forwarded to `brod_utils:resolve_offset/5`

zmstone: i.e. call `brod_utils:resolve_offset(Hosts, Topic, Partition, Time, ConnCfg).`

test/brod_consumer_SUITE.erl
```diff
@@ -362,6 +363,23 @@ t_direct_fetch(Config) when is_list(Config) ->
...
+  Client = ?config(client),
+  Topic = ?TOPIC,
+  Partition = 0,
+  Key = make_unique_key(),
+  Value = <<>>,
+  ClientConfig = lists:merge(?config(client_config), 
+                              [{connect_timeout, timer:seconds(45)}]),
+  ok = brod:produce_sync(Client, ?TOPIC, Partition, Key, Value),
+  {ok, Offset} = brod:resolve_offset(?HOSTS, Topic, Partition,
+                                     ?OFFSET_LATEST, ClientConfig),
```
zmstone: I mean test `brod:resolve_offset/6`

escobera: Ok, I'll do that

test/brod_consumer_SUITE.erl
```diff
@@ -362,6 +363,22 @@ t_direct_fetch(Config) when is_list(Config) ->
   ?assertEqual(Key, Msg#kafka_message.key),
   ok.
 
+t_direct_fetch_with_connect_timeout(Config) when is_list(Config) ->
```
zmstone: one last thing
this test case is duplicating the above one.
Maybe rename it to `t_resolve_offset` and no need to test anything else than the call to `brod:resolve_offset/6`.
e.g.
```
t_resolve_offset(Config) when is_list(Config) ->
  Client = ?config(client),
  Topic = ?TOPIC,
  Partition = 0,
  {ok, Offset} = brod:resolve_offset(?HOSTS, Topic, Partition,
                                     ?OFFSET_LATEST, ?config(client_config),
                                     #{timeout => timer:seconds(45)}),
  ?assertTrue(is_integer(Offset)).
```

src/brod.erl
```diff
@@ -850,7 +851,15 @@ resolve_offset(Hosts, Topic, Partition, Time) ->
                      offset_time(), conn_config()) ->
         {ok, offset()} | {error, any()}.
 resolve_offset(Hosts, Topic, Partition, Time, ConnCfg) ->
-  brod_utils:resolve_offset(Hosts, Topic, Partition, Time, ConnCfg).
+ brod_utils:resolve_offset(Hosts, Topic, Partition, Time, ConnCfg).
```
zmstone: indent two spaces.

escobera:
@zmstone ok I think I fixed the problems you pointed. 

I'm sorry for this back and forth, it is my first contribution to an Erlang project.

Anything else you think is laking in this PR just let me know.

zmstone (approved):
maybe add a record in changelog.md
Looks good otherwise.
Ping @mikpe @robertoaloi for (squash) merge

escobera:
Any news here guys? If I need to do anything else just lemme know!

Cc: @mikpe @robertoaloi @zmstone

mikpe (approved):

mikpe:
Merged, added changelog entry, tagged as 3.12.0.

PR #396, opened 2020-05-11
fix aborted transactions

Leesanwei:
We use Azure event hub and we get :undefined for aborted_transactions, this could help us


zmstone:
duplicates #392

PR #403, opened 2020-05-27
Fix brod:create_topics typespecs

DianaOlympos:
I updated the typespec to reflect what it does today. I am not sure it should be the behaviour and i did not update the documentation as i do not fully understand why it does not always return the struct as expected.


Coffei:
This problem seems to concern `delete_topics` APIs as well. Could you also fix that?

DianaOlympos:
Done. Tbh i think it would be better if we understood where it come from because i am not sure we can really receive a `{ok, kpro:struct()}` anymore. but not 100% sure.

k32:
Hi @DianaOlympos,

Thanks for the PR!

k32 (approved):
New specs look more correct when I compare them with the typer output.

PR #405, opened 2020-06-01
brod_cli_tests:cmd/1: ensure brod_kafka_apis is gone before returning

mikpe:
A pending PR has a build error with OTP-19, caused by what looks like a race between shutting down the brod application and its linked processes for one test, and starting it for another: the start failed because brod_kafka_apis hadn't terminated yet.

This adds code to brod_cli_tests:cmd/1 to also ensure that brod_kafka_apis is gone before returning, if necessary by killing it and looping briefly.

See: https://travis-ci.org/github/klarna/brod/jobs/692218660

### DIFF
--- a/test/brod_cli_tests.erl
+++ b/test/brod_cli_tests.erl
@@ -193,7 +193,20 @@ cmd(Args) ->
     catch brod:stop_client(brod_cli_client),
     Result
   after
-    _ = brod:stop()
+    _ = brod:stop(),
+    ensure_down(brod_kafka_apis)
+  end.
+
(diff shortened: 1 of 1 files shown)


mikpe:
WDYT @zmstone?  This is a bit of a stab in the dark ...

qzhuyan (approved):

zmstone:
the process `brod_kafka_apis` is  linked to root supervisor but not managed as a child.
maybe we should supervise it so it can be stopped properly.
Or move the wait to `brod:stop/1`

mikpe:
We'll improve the fix in a follow-up, for now this is a safe and sufficient change.

PR #409, opened 2020-06-15
Topic subscriber owns consumers

k32:
Note: this is not a BWC change, since `register_self` is `false` by default, so any code that used `brod:get_consumer` for some reason will be broken.


mikpe:
I'll take a look tomorrow, but I'm concerned about the non backwards-compatible nature of this change.

src/brod_topic_subscriber.erl
```diff
@@ -94,21 +94,22 @@
...
         }).
 
 -type state() :: #state{}.
 
 %% delay 2 seconds retry the failed subscription to partiton consumer process
 -define(RESUBSCRIBE_DELAY, 2000).
 
--define(LO_CMD_START_CONSUMER(ConsumerConfig, CommittedOffsets, Partitions),
+-define(LO_CMD_START_CONSUMER(CommittedOffsets, Partitions),
         {'$start_consumer', ConsumerConfig, CommittedOffsets, Partitions}).
```
mikpe, line 149: The body of this macro now has a free variable.

mikpe (commented):
I would strongly prefer not to break existing users in a non-major version bump.  I.e., the new behaviour ought to be opt-in (which should be good enough for us), and then _perhaps_ made mandatory in a subsequent major version bump.

MichalDolata:
Hi @k32! What is the state of this PR? Do you need any help to make it merge ready?

k32:
Hi,

Sadly, I no longer work on brod. As far as I remember, this PR was in a working shape overall, but it made ownership and the supervision tree rather messy, hence I was hesitant to merge it.

PR #410, opened 2020-06-16
Group subscriber v2 terminate

k32:
This PR introduces an optional `brod_group_subscriber_v2` `terminate` callback.

### DIFF
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -70,9 +70,13 @@
         ]).
 
 %% Subscriber API
+-export([ start_link_group_subscriber_v2/1
+        , start_link_topic_subscriber/1
+        ]).
+
+%% Deprecated API
 -export([ start_link_group_subscriber/7
         , start_link_group_subscriber/8
-        , start_link_group_subscriber_v2/1
         , start_link_topic_subscriber/5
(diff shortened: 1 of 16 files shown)


mikpe (approved):

src/brod.erl
```diff
@@ -70,9 +70,13 @@
         ]).
 
 %% Subscriber API
+-export([ start_link_group_subscriber_v2/1
```
mikpe, line 73: nit: it's unfortunate that there's a _v2 in the name as the /1 arity makes it unambiguous.  Maybe make the _v2 function a compat alias for start_link_group_subscriber/1, then drop the _v2 one when the old API is dropped?

mikpe (approved):

src/brod.erl
k32: Not sure if I follow you here: `brod_group_subscriber_v2:start_link/1` is an existing API that was introduced together with `brod_group_subsriber_v2` behavior, so renaming it won't be BWC. I didn't add a similar new API for the old `brod_group_subscriber` module, because I consider it problematic, and I hope to get it removed in the next major release (and rename `brod_group_subscriber_v2` module into `brod_group_subscriber`).

PR #421, opened 2020-10-05
Fix default timeout macro

mikpe:


jesperes (approved):
Thanks for adding the missing 3.15.0 changelog entry!

qzhuyan (approved):

PR #425, opened 2020-10-12
Add elixir example

prabello:
This introduces a folder `./contrib/examples` to add code examples of `brod` usage
A small tutorial on how to configure and use `brod` with Elixir directly, without depending on other libraries

As previously mentioned on: https://github.com/klarna/brod/issues/414

I had the intentions of working more on this, but since I was unable I think it makes more sense to increase it over time with some feedback, please let me know if anything is unclear, so I can fix it.


mikpe (approved):

prabello:
Any way i can re-run the CI?

mikpe (approved):

PR #428, opened 2020-11-30
Use GitHub actions

jesperes:
Following Klarna guidelines for open source projects, this PR sets up a CI workflow for GitHub Actions.


jesperes:
GitHub Actions build for commit bf36b86 is green, but the Travis build fails because there are lots of Elvis failures which nobody seems to care about.

mikpe (approved):

PR #435, opened 2021-02-23
Update Copyright Year

robertoaloi:


NOTICE
```diff
@@ -1,5 +1,5 @@
 Kafka client library in Erlang
-Copyright 2014, 2015 Klarna AB
+Copyright 2014-2021 Klarna AB
 
 This product includes software developed by
 Klarna AB (https://www.klarna.com)
```
mikpe: Please change these two occurrences of "Klarna AB" to say "Klarna Bank AB" like the rest of the files do.

robertoaloi: Fixed.

mikpe (approved):

PR #439, opened 2021-02-26
Fix brod_group_subscriber_v2 crash on shutdown

### LINKED ISSUE #438: Address remarks
@epsylonix commented on this pull request.

Hi @k32, could you please look into these probable issues?

In src/brod_group_subscriber_v2.erl:

>      [TopicPartition|_] ->
-      handle_worker_failure(TopicPartition, Pid, Reason, State)
+      handle_worker_failure(TopicPartition, Pid, Reason, State);
+    _ -> % Other process wants to kill us, supervisor?
+      {stop, shutdown}

this should probably be {stop, shutdown, State}

In src/brod_group_subscriber_v2.erl:

>      [TopicPartition|_] ->
-      handle_worker_failure(TopicPartition, Pid, Reason, State)
+      handle_worker_failure(TopicPartition, Pid, Reason, State);
+    _ -> % Other process wants to kill us, supervisor?

I believe there is a bug here: when a worker is stopped (in terminate_worker), brod_group_subscriber_v2 removes its partition => worker_pid from the state but since it is trapping exits, brod_group_subscriber_v2 will receive a normal exit from that worker later, and then this clause will catch this exit and brod_group_subscriber_v2 will shut down. I believe at this moment any assignments_revoked call would lead to brod_group_subscriber_v2 crash and restart.

There is possibly another issue with the workers termination process:
brod_topic_subscriber is stopped by casting stop and waiting for the 'DOWNmessage (https://github.com/kafka4beam/brod/blob/master/src/brod_topic_subscriber.erl#L256), butbrod_group_subscriber_v2is trapping exits, so it would receiveEXITinstead and noDOWN` message will be sent if I'm not mistaken

epsylonix:
This PR fixes #438

I believe there is still one possible issue with a termination process: `brod_topic_subscriber:stop` uses `cast(Pid, stop)` + infinite `receive`, but since `brod_topic_subscriber` makes calls to the client module (i.e. `CbModule:handle_message`), it is possible for those calls to hang indefinitely and that would prevent it from processing the `stop` message. Maybe it would make sense to wait for `brod_topic_subscriber`'s graceful shutdown only for so long and then give up and shut it down forcefully. This is not addressed  here.

### DIFF
--- a/src/brod_group_subscriber_v2.erl
+++ b/src/brod_group_subscriber_v2.erl
@@ -365,7 +365,8 @@ handle_info({'EXIT', Pid, Reason}, State) ->
     [TopicPartition|_] ->
       handle_worker_failure(TopicPartition, Pid, Reason, State);
     _ -> % Other process wants to kill us, supervisor?
-      {stop, shutdown}
+      ?BROD_LOG_INFO("Received EXIT:~p from ~p, shutting down", [Reason, Pid]),
+      {stop, shutdown, State}
   end;
 handle_info(_Info, State) ->
   {noreply, State}.
--- a/src/brod_group_subscriber_v2.erl
+++ b/src/brod_group_subscriber_v2.erl
@@ -416,6 +416,7 @@ terminate_all_workers(Workers) ->
 terminate_worker(WorkerPid) ->
   case is_process_alive(WorkerPid) of
     true ->
+      unlink(WorkerPid),
       brod_topic_subscriber:stop(WorkerPid);
     false ->
       ok
--- a/test/brod_group_subscriber_SUITE.erl
+++ b/test/brod_group_subscriber_SUITE.erl
@@ -52,6 +52,7 @@
         , t_assign_partitions_handles_updating_state/1
         , v2_coordinator_crash/1
         , v2_subscriber_shutdown/1
+        , v2_subscriber_assignments_revoked/1
         ]).
 
 -define(MESSAGE_TIMEOUT, 30000).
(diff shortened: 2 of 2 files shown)


src/brod_group_subscriber_v2.erl
```diff
@@ -415,6 +416,7 @@ terminate_all_workers(Workers) ->
 terminate_worker(WorkerPid) ->
   case is_process_alive(WorkerPid) of
     true ->
+      unlink(WorkerPid),
       brod_topic_subscriber:stop(WorkerPid);
```
k32, line 420: This `unlink` call could be moved inside `brod_topic_subscriber:stop` to fix this issue in a generic way.

k32:
Thanks for the PR, I have a small remark

k32 (approved):

src/brod_group_subscriber_v2.erl
epsylonix: Thanks for the review! Sure, `unlink` could be moved to `stop`, but I'm not sure it is a desirable property in a general case - it is a little less explicit and in some cases the user might not even want that - for example, some other `brod_topic_subscriber`'s user may want to handle a crash during `brod_topic_subscriber` termination and `unlink`'ing would prevent that. Another argument I have is that since `brod_group_subscriber_v2` traps exits, it has to make sure it handles them correctly so it shouldn't probably depend on a `stop`s side effect for that. For consistency sake if this change is made then it would make sense to do the same in the other modules' `stop`'s too by the way.

But I might be wrong here, it's your call :)

k32: I think stopping of all processes should be done uniformly, but now it's not the case for different reasons. I guess we can fix it later.

PR #444, opened 2021-03-24
Do not try to format the buffer state when the process abnormally terminates and logs the error

### LINKED ISSUE #443: eheap_alloc error when calling exit in brod_producer_buffer.erl
We found that when kafka is down and `brod_producer.erl` reaches the `reached_max_retries`, this [exit](https://github.com/kafka4beam/brod/blob/master/src/brod_producer_buffer.erl#L304) does  the VM to crash with a eheap_alloc message. The error message shows that the request contains a disproportional number of bytes like`eheap_alloc: Cannot allocate 863825524200 bytes of memory (of type"message").`

When `brod_producer` exits, `gen_server` tries to format the `#state{}` with the buffer data. This can contain thousand of elements and does `gen_server:terminate` to request more memory and crash. 

A solution we found is to create a custom [format_status function](http://erlang.org/doc/man/gen_server.html#Module:format_status-2) in `brod_producer.erl` that replaces the buffer in the state when gen_server call it [here](https://github.com/erlang/otp/blob/master/lib/stdlib/src/gen_server.erl#L1246) with `terminate` Opt.

gonzalobf:
Fixes #443. ~~I used `opaque_buffer` atom so it appears in the log and the reader can know why the buffer is not serialized but I'm happy to change it.~~ 
I replaced `opaque_buffer` with a buffer where the list and the queues are empty

### DIFF
--- a/src/brod_producer.erl
+++ b/src/brod_producer.erl
@@ -31,6 +31,7 @@
         , handle_info/2
         , init/1
         , terminate/2
+        , format_status/2
         ]).
 
 -export_type([ config/0 ]).
@@ -423,6 +424,15 @@ terminate(Reason, #state{client_pid = ClientPid
       ok
   end,
   ok.
+
+%% @private
+format_status(Opt, [_PDict, State]) ->
+  %% Do not format the buffer attibute when process terminates abnormally and logs an error
+  %% but allow it when is a sys:get_status/1.2
+  case Opt of
+      terminate -> State#state{buffer=opaque_buffer};
+      _ -> [{data, [{"State", State}]}]
+  end.
 
 %%%_* Internal Functions =======================================================
 
--- a/src/brod_producer.erl
+++ b/src/brod_producer.erl
@@ -426,11 +426,14 @@ terminate(Reason, #state{client_pid = ClientPid
   ok.
 
 %% @private
-format_status(Opt, [_PDict, State]) ->
+format_status(Opt, [_PDict, State=#state{buffer = Buffer}]) ->
   %% Do not format the buffer attibute when process terminates abnormally and logs an error
   %% but allow it when is a sys:get_status/1.2
   case Opt of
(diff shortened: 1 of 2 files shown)


src/brod_producer.erl
```diff
@@ -424,6 +425,18 @@ terminate(Reason, #state{client_pid = ClientPid
   end,
   ok.
 
+%% @private
+format_status(Opt, [_PDict, State=#state{buffer = Buffer}]) ->
+  %% Do not format the buffer attibute when process terminates abnormally and logs an error
```
mikpe: Nit: s/attibute/attribute/

src/brod_producer_buffer.erl
```diff
@@ -364,6 +365,11 @@ now_ms() ->
   {M, S, Micro} = os:timestamp(),
   ((M * 1000000) + S) * 1000 + Micro div 1000.
 
+-spec create_opaque_buffer(buf()) -> buf().
+create_opaque_buffer(Buffer = #buf{}) ->
```
mikpe: Nit: The naming is a bit off, it's not really creating something opaque but simply _emptying_ the buffer, compare with the `is_empty/1` predicate.

src/brod_producer.erl
```diff
@@ -424,6 +425,18 @@ terminate(Reason, #state{client_pid = ClientPid
...
+  %% but allow it when is a sys:get_status/1.2
+  case Opt of
+      terminate ->
```
pablocostass: nit: this `case` statement could be very well done with function clauses

zmstone (approved):
Thank you for the PR.
Will tag a release after another minor update PR in changelog.md.

gonzalobf:
@zmstone Thank you :)

zmstone:
having trouble publish to hex. someone please help.

zmstone:
tagged 3.15.2 because 3.15.1 is taken already.

PR #447, opened 2021-03-30
improve(subscriber_v2): flush offset commits before terminate

### LINKED ISSUE #442: Graceful shutdown
Hi guys. 
I think I'm doing something wrong, seems that I don't get the supervision strategy.
I have a pipeline that parses and sends all messages to a new topic. 
When I shut down the server, some messages are sent but not commited, which causes duplication (I need exactly once delivery).
Here is an example code (I'm sorry, it's in elixir, but could be helpful for other folks).

```elixir
defmodule TestPipeline do
  @behaviour :brod_group_subscriber_v2
  @source :test_consumer

  require Logger

  def child_spec(_opts) do
    %{
      id: __MODULE__,
      start: {__MODULE__, :start, []},
      type: :worker,
      restart: :permanent,
      shutdown: 10_000
    }
  end

  def start() do
    group_config = [
      offset_commit_policy: :commit_to_kafka_v2,
      offset_commit_interval_seconds: 2,
      rejoin_delay_seconds: 3,
      reconnect_cool_down_seconds: 10
    ]

    config = %{
      client: @source,
      group_id: "test",
      topics: ["test"],
      cb_module: __MODULE__,
      group_config: group_config,
      consumer_config: [begin_offset: :earliest]
    }

    :brod.start_link_group_subscriber_v2(config)
  end

  def init(_arg, _arg2) do
    {:ok, []}
  end

  def handle_message(messages, _state) do
    IO.inspect("processing message #{inspect(messages)}") # here I send the same batch to the new topic.
    :timer.sleep(3_000)
    IO.inspect("commiting offset")
    {:ok, :commit, nil}
  end
end
```

and my test:
```elixir
  test "shutdown gracefully" do
    brokers = [{:kafka, 9092}]

    :ok = :brod.start_client(brokers, :test_consumer)

    {:ok, pid} = start_supervised(TestPipeline)
    :timer.sleep(1_000)
    :init.stop()
    IO.inspect("shutting down")
    :timer.sleep(20_000)
  end
```

And it shows:
```
"processing message .... (always same messages)"
"shutting down"
```
But the offset is not commited

Thanks!

zmstone:
fixes #442

### DIFF
--- a/src/brod_group_subscriber_v2.erl
+++ b/src/brod_group_subscriber_v2.erl
@@ -113,7 +113,7 @@
         { config                  :: subscriber_config()
         , message_type            :: message | message_set
         , group_id                :: brod:group_id()
-        , coordinator             :: pid()
+        , coordinator             :: undefined | pid()
         , generation_id           :: integer() | ?undef
         , workers = #{}           :: workers()
         , committed_offsets = #{} :: committed_offsets()
@@ -357,12 +357,11 @@ handle_cast(_Cast, State) ->
 %% Handling all non call/cast messages
 %% @end
 %%--------------------------------------------------------------------
+handle_info({'EXIT', Pid, _Reason}, #state{coordinator = Pid} = State) ->
+    {stop, {shutdown, coordinator_failure}, State#state{coordinator = undefined}};
 handle_info({'EXIT', Pid, Reason}, State) ->
-  L = [TP || {TP, Pid1} <- maps:to_list(State#state.workers), Pid1 =:= Pid],
-  case L of
-    _ when Pid =:= State#state.coordinator ->
-      error(coordinator_failure);
-    [TopicPartition|_] ->
+  case [TP || {TP, Pid1} <- maps:to_list(State#state.workers), Pid1 =:= Pid] of
+    [TopicPartition | _] ->
       handle_worker_failure(TopicPartition, Pid, Reason, State);
     _ -> % Other process wants to kill us, supervisor?
       ?BROD_LOG_INFO("Received EXIT:~p from ~p, shutting down", [Reason, Pid]),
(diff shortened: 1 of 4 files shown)


k32 (approved):

zmstone:
tests to be added.

src/brod_group_subscriber_v2.erl
```diff
@@ -379,14 +378,29 @@ handle_info(_Info, State) ->
...
 
+%% best-effort commits flush, this is a synced call,
+%% worst case scenario, it may timeout after 5 seconds.
+flush_offset_commits(GroupId, Coordinator) when is_pid(Coordinator) ->
+    case brod_group_coordinator:commit_offsets(Coordinator) of
+        ok -> ok;
+        {error, Reason} ->
+            ?BROD_LOG_ERROR("group_subscriber_v2 ~s failed to flush commits "
+                            "before termination", [GroupId, Reason])
+    end;
```
mikpe, line 401: Nit: looks like you have deeper indentation here than in the rest of the module.
LGTM otherwise.

zmstone:
rebased. last two commits are new.

src/brod_group_subscriber_v2.erl
```diff
@@ -379,28 +379,38 @@ handle_info(_Info, State) ->
...
+        ok -> ok;
+        {error, Reason} ->
+            ?BROD_LOG_ERROR("group_subscriber_v2 ~s failed to flush commits "
+                            "before termination ~p", [GroupId, Reason])
+    end;
+flush_offset_commits(_, _) ->
+    ok.
+
 -spec handle_worker_failure(brod:topic_partition(), pid(), term(), state()) ->
                                no_return().
```
mikpe: nit: this function no longer calls exit/1 so the -spec is a bit off

mikpe (approved):

src/brod_group_subscriber_v2.erl
zmstone: thank you @mikpe 
Fixed it and squashed to the second commit.

PR #451, opened 2021-04-24
fix: no link-start for temp connections

### LINKED ISSUE #450: Worker terminating with reason `:normal` after calling `resolve_offset`
We're using elsa as a wrapper for brod in our Elixir-project. We are facing unexpected behavior when calling brod's `resolve_offset` function. That is, our elsa consumer that operates on the same topic receives an exit signal with reason `:normal`. The exit signal arrives not immediately, but a short time after the call of `resolve_offset`. The consumer works perfectly fine, if we don't call `resolve_offset` on the same topic. 

We tried to narrow down the reason for this and it seems the exit signal is originated somewhere in brod. We weren't able to follow the signal chain entirely, so take this with a grain of salt. 

One thing we noticed is that the following changes in the parameters of `with_conn` stop the unexpected behavior and our consumer doesn't quit after calling `resolve_offset` on the same topic:

https://github.com/kafka4beam/brod/blob/master/src/brod_utils.erl#L156

    %% @doc Resolve timestamp to real offset.
    -spec resolve_offset([endpoint()], topic(), partition(),
                         offset_time(), conn_config(),
                        #{timeout => kpro:int32()}) ->
            {ok, offset()} | {error, any()}.
    resolve_offset(Hosts, Topic, Partition, Time, ConnCfg, Opts) ->
      with_conn(
        % we changed this, as seen in get_metadata
        % kpro:connect_partition_leader(Hosts, ConnCfg, Topic, Partition, Opts),
        Hosts, ConnCfg,
        fun(Pid) -> resolve_offset(Pid, Topic, Partition, Time) end).

Do you have any idea if this is a bug or are we just doing it wrong? 
We really appreciate any hints, since we've already spent days debugging this issue.

zmstone:
fixes #450

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,8 @@
 * TBD
   *
+* 3.15.4
+  * Avoid start_link for temp connection usages
+    affected APIs: fetch_committed_offsets, fetch, resolve_offset, create_topics, delete_topics
 * 3.15.3
   * Try to commit acknowledged offsets when brod_group_subscriber_v2 terminates
   * Fix process leak, close connection after offset is fetched in brod_utils:fetch_committed_offsets/3
--- a/src/brod_utils.erl
+++ b/src/brod_utils.erl
@@ -93,7 +93,7 @@ create_topics(Hosts, TopicConfigs, RequestConfigs) ->
                     validate_only => boolean()}, conn_config()) ->
         {ok, kpro:struct()} | {error, any()} | ok.
 create_topics(Hosts, TopicConfigs, RequestConfigs, ConnCfg) ->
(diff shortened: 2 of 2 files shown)


changelog.md
```diff
@@ -1,5 +1,8 @@
 * TBD
   *
+* 3.15.4
+  * Avoid start_lnik for temp connection usages
```
mikpe: nit: s/lnik/link/

mikpe (approved):

changelog.md
zmstone: thank you. squashed the fix commit.

zmstone:
tests are passing. failed to send coverage report to coveralls. seem to be temporary issue.

PR #452, opened 2021-04-26
Fix the exponential growth of the producer buffer

dszoboszlay:
Referencing `State#state.partition` from within the `BufCb` fun adds the entire state (including the producer buffer) to the free variables of the fun. Then the fun is added to the producer buffer and in turn to the new version of the state.

This means that when multiple produce requests are buffered, the "flattened size" of the producer buffer will grow exponentially. The observable memory growth of the producer process will be linear, because the old and the new state reference a lot of common terms on the heap. However, when the producer buffer is copied (e.g. when sending it in a message) all those duplicate references lead to creating duplicate copies of the referenced terms, hence the memory need of the message will grow exponentially.

This problem can be easily avoided by making sure only the necessary `partition` field of the state becomes a free variable of the fun.

### DIFF
--- a/src/brod_producer.erl
+++ b/src/brod_producer.erl
@@ -345,7 +345,7 @@ handle_info({'DOWN', _MonitorRef, process, Pid, Reason},
       {ok, NewState} = schedule_retry(State#state{buffer = Buffer}),
       {noreply, NewState#state{connection = ?undef, conn_mref = ?undef}}
   end;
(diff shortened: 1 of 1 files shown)


zmstone (approved):

zmstone:
will create a release later.

zmstone:
released 3.15.5

PR #455, opened 2021-04-30
Update field names everywhere to make tests pass

tpitale:
cc @zmstone I've got all of the unit tests passing with `make ut`. I'm not sure what else might need updating. You mentioned something about the latest APIs, as well.


tpitale:
Related to discussion #454

zmstone:
Kafka support API version range query right after a connection is established.
`kafka_protocol` queries Kafka's API versions, and intersects the ranges with its own supported ranges.
 in order to avoid surprises, `brod` also has the supported ranges [hard-coded here](
https://github.com/kafka4beam/brod/blob/582f95cf01a1d72334078f0a31b6e6fd22f2051d/src/brod_kafka_apis.erl#L140-L158)

let's focus on the data structure changes first,
and maybe send another PR to update the APIs.

PR #460, opened 2021-06-04
Declosurize

mikpe:
Brod uses a lot of long-lived anonymous functions, which cause problems during live upgrades. In our case a change in build procedure caused the `brod` .beams to change unexpectedly, causing a lot of `badfun` errors after the upgrade. (If you know `brod` will be upgraded you can stop and restart it, but in this case we didn't know.)

To avoid these issues this changes several callback functions to accept a `{Fun, Arg}` representation. At the call sites, `Arg` is prepended to the list of actual parameters. This allows us to use exported functions, `fun Module:Fun/N`, as the callbacks, making the code live-upgrade safe.

In two cases the code wrapped data in functions to hide it from crash logs, but that isn't necessary any more.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-* TBD
+* 3.15.6
   * Eliminate long-lived anonymous function closures
 * 3.15.5
   * Fix exponential growth of `brod_producer` buffer
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 * TBD
+  * Eliminate long-lived anonymous function closures
 * 3.15.5
   * Fix exponential growth of `brod_producer` buffer
 * 3.15.4
--- a/src/brod_producer.erl
+++ b/src/brod_producer.erl
@@ -35,6 +35,7 @@
         ]).
 
 -export([ do_send_fun/4
+        , do_no_ack/2
         ]).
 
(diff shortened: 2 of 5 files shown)


mikpe:
Looks like rebar3 is broken in your OTP-21 job. I have no idea why the OTP-22/23 jobs were cancelled.

zmstone:
> Looks like rebar3 is broken in your OTP-21 job. I have no idea why the OTP-22/23 jobs were cancelled.

All jobs in a matrix are cancelled if any one of them fails (it's a github action feature).

It's likely the default rebar3 for github action has been upgraded to one built on OTP-22.
I have sent https://github.com/kafka4beam/brod/pull/461 to fix it.

zmstone:
@mikpe if you fetch and rebase, the ci should be able to run.

mikpe:
Rebased and added a note to the changelog.

zmstone (approved):
thank you @mikpe 
Please change TBD to 3.15.6 if you want a tag

mikpe:
> thank you @mikpe
> Please change TBD to 3.15.6 if you want a tag

Done, thanks.

mikpe:
Ping. Can someone merge & tag please.

mikpe:
> 3.15.6 tagged

Thanks

PR #465, opened 2021-07-19
Add join_group rebalance_timeout_ms field

erszcz:
When using [Vectorized Redpanda](https://vectorized.io/redpanda/) as the Kafka broker we've run into an issue with consumer group stabilisation. My research has lead to the `join_group` request not carrying `rebalance_timeout_ms` value, which in turn lead to [`rebalance_timeout()` computation on Redpanda side returning -1](https://github.com/vectorizedio/redpanda/blob/a8d19c32c95c2c313e7eea69defc058a8b269e2b/src/v/kafka/server/group.cc#L251-L264). Each time a consumer wanted to join an existing group a rebalance would be started, but with the timeout value of -1 existing members wouldn't have time to rejoin. This would lead to the kicked out members retrying to join the group, which in turn would rebalance again and kick out previously joining members. This would loop indefinitely.

With this fix applied, Redpanda doesn't use -1 as the group join timeout, therefore the group stabilises as expected. I've contacted them about considering using a more reasonable default, if `join_group` request doesn't carry `rebalance_timeout_ms`.

In the meantime, I think this fix might be useful to other Brod users.

### DIFF
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -49,7 +49,7 @@
 
 %% default configs
 -define(SESSION_TIMEOUT_SECONDS, 30).
--define(REBALANCE_TIMEOUT_SECONDS, 3).
+-define(REBALANCE_TIMEOUT_SECONDS, 2).
 -define(HEARTBEAT_RATE_SECONDS, 5).
 -define(PROTOCOL_TYPE, <<"consumer">>).
 -define(MAX_REJOIN_ATTEMPTS, 5).
@@ -135,6 +135,7 @@
           %% configs, see start_link/5 doc for details
         , partition_assignment_strategy  :: partition_assignment_strategy()
         , session_timeout_seconds        :: pos_integer()
+        , rebalance_timeout_seconds      :: pos_integer()
         , heartbeat_rate_seconds         :: pos_integer()
         , max_rejoin_attempts            :: non_neg_integer()
         , rejoin_delay_seconds           :: non_neg_integer()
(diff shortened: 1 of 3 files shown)


src/brod_group_coordinator.erl
```diff
@@ -49,6 +49,7 @@
 
 %% default configs
 -define(SESSION_TIMEOUT_SECONDS, 30).
+-define(REBALANCE_TIMEOUT_SECONDS, 2).
```
zmstone: To make it backward compatible, please default to `SESSION_TIMEOUT_SECONDS` here.
If I remember correctly, this is how newer version Kafka handles version `0` join request:
it takes the session timeout for rebalance timeout.

erszcz: Thanks, will do :+1:

src/brod_group_coordinator.erl
```diff
@@ -180,15 +182,23 @@
...
 %%      A group member may also consider the coordinator broker 'down' if no
 %%      heartbeat response response received in the past N seconds.</li>
 %%
-%%  <li>`heartbeat_rate_seconds' (optional, default = 2)
+%%
+%%  <li>`rebalance_timeout_seconds' (optional, default = 30)
+%%
+%%      Time in seconds for each worker to join the group once a rebalance
+%%      has begun. If the timeout is exceeded, then the worker will be
+%%      removed from the group, which will cause offset commit failures.
```
mikpe: Missing `</li>`.

mikpe (approved):

zmstone:
@erszcz thank you for the PR.
If you don't mind, I'll squash-merge this PR.

erszcz:
@zmstone Sure, no problem with a squash-merge. Thanks for the review!

zmstone:
tagged 3.16.1

zmstone:
also published to hex

erszcz:
Awesome, thank you!

PR #467, opened 2021-07-27
Fix typo

george124816:
a typo on brod log error
paritition to partition


zmstone (approved):

zmstone:
thank you for the fix.

PR #495, opened 2022-04-04
Fix spec for brod_utils:delete_topics

hrubi:


src/brod_utils.erl
```diff
@@ -103,14 +103,14 @@ create_topics(Hosts, TopicConfigs, RequestConfigs, ConnCfg) ->
             end).
 %% @equiv delete_topics(Hosts, Topics, Timeout, [])
 -spec delete_topics([endpoint()], [topic()], pos_integer()) ->
-        {ok, kpro:struct()} | {error, any()}.
+        ok | {ok, kpro:struct()} | {error, any()}.
```
zmstone: In which case does it return an ok tuple?

hrubi: In none. It always returns either `ok` or the error tuple due to this:
```
parse(delete_topics, _, Msg) ->
  ok = throw_error_code(kpro:find(responses, Msg));
```
I will update the spec both here and in `brod:delete_topics`.

zmstone:
@hrubi thank you for the PR.
could you help to update the changelog.md ? add a new version at the top (3.16.3)

hrubi:
@zmstone sure. I've updated the changelog.

PR #499, opened 2022-05-01
Use ex_doc to generate docs

spencerdcarlson:
Proposal to use [rebar3_hex](https://hexdocs.pm/rebar3_hex/readme.html) and [ex_doc](https://hex.pm/packages/ex_doc) to generate docs


spencerdcarlson:
One thing I've seen users get tripped up on is when switching from using `brod_group_subscriber` to `brod_group_subscriber_v2` the the return type of `{ok, ack, cb_state()}` means very different things when implementing `handle_message/4` vs `handle_message/2`. It would be nice if we could clarify this in the docs. Unfortunately it looks like erlang's edoc is lacking the ability to add any useful docs (besides the function name and arity).

Using ex_docs at least displays the input types and return types. Ex_docs has the ability to publish guides or examples.

## handle_message/4
handle_message-4

## handle_message/2
handle_message-2

## Auth Guides Example

auth-guide

zmstone:
Hi @spencerdcarlson 
Possible to make it work on OTP 23?
It's fine not to build docs on OTP 23, just need github actions to pass the tests.

guides/examples/Authentication.md
```diff
@@ -0,0 +1,102 @@
...
+           sasl: {
+             :plain,
+             System.get_env("KAFKA_USERNAME"),
+             System.get_env("KAFKA_PASSWORD")
+           }
+         ]
+       ]
+```
+
+The examples above are use `/etc/ssl/certs/ca-certificates.crt` which is the certificate authority that comes
```
zmstone:
```suggestion
The examples above are using `/etc/ssl/certs/ca-certificates.crt` which is the certificate authority that comes
```

guides/examples/elixir/Consumer.md
```diff
@@ -0,0 +1,71 @@
...
+The example module implements `child_spec/1` so that our consumer can be started by a Supervisor. The restart policy is set to `:temporary`
+because, in this case, if a message can not be processed, then there is no point in restarting. This might not always
+be the case.
+
+See `brod_group_subscriber_v2:start_link/1` for details on the configuration options.
+
+See docs for more details about the required or optional [callbacks](file:///Users/sdc/code/sdc/brod/doc/brod_group_subscriber_v2.html#callbacks)
+
+## Warning
+If you are not running your application in cluster mode you may go into some issues as the `group_subscriber` on multiple nodes may force each other re-join the group.
```
zmstone: Sorry, I’m not quite following this warning.
A consumer group member joining/leaving always  forces others to rejoin.
What does this “cluster mode” mean here?

guides/examples/elixir/Publisher.md
```diff
@@ -0,0 +1,111 @@
...
+![kafdrop](../../../contrib/examples/elixir/docs/kafdrop.png)
+
+We can see that there is something on partition 0
+
+Opening it up we see
+![topic](../../../contrib/examples/elixir/docs/streamingevents.png)
+
+### Using partition key
+
+TODO: Update to talk about `:hash` as partitioner
```
zmstone: Is this TODO still relevant?
Hash partitioner is already talked about below?

guides/examples/elixir/Publisher.md
```diff
@@ -0,0 +1,111 @@
...
+
+The most common way to send messages to kafka is by using a partition key and based on that deciding to what partition the message should go, let's see how we can achieve that
+
+First, we need to know how many partitions our topic have, so we don't try sending the message to a non-existing partition, for that we can also use brod
```
zmstone: We should perhaps give the readers a note that this is how brod maps key to partition number internally. Not something for one using brod to implement?

zmstone:
Awesome @spencerdcarlson 
A few nitpicking comments hope you don’t mind

guides/examples/elixir/Consumer.md
spencerdcarlson: I will remove this. I copied this file from the existing elixir example [README](https://github.com/kafka4beam/brod/blob/master/contrib/examples/elixir/README.md). I am _guessing_ they are referring to Kafka's consumer rebalancing protocol, but I don't know how that relates to clustered mode (assuming "clustered mode" is Erlang Distributed mode)

guides/examples/elixir/Publisher.md
spencerdcarlson: Same as above

guides/examples/elixir/Publisher.md
spencerdcarlson: I totally agree. We should just call out that `brod:produce_sync/3` and the like do this. And _maybe_ give an example of a custom partition routing strategy.

zmstone:
Looks good to me.
Maybe rebase or squash to tidy up the commits ?

spencerdcarlson:
It won't let me squash locally. I thought there was an option to squash when merging.

PR #501, opened 2022-05-10
Move Dialyzer & Xref. Make CT logs env specific

spencerdcarlson:
[This](https://github.com/kafka4beam/brod/runs/6368748097?check_suite_focus=true) single failed test prompted this MR
* The latest test run overwrites the any previous CT logs. This should create an artifact for each test matrix.  
* We only need to run Dialyzer and Xref once. Adding it to the Lint job.
* We can cache the entire `_build` directory because `rebar3 compile` will compile any changed files

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches:
       - master
+env:
+  OTP_VERSION: '24.1'
+  REBAR_VERSION: '3.17.0'
 
 jobs:
   lint:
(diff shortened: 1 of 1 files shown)


zmstone (approved):

.github/workflows/build.yml
```diff
@@ -24,11 +24,17 @@ jobs:
...
         uses: actions/cache@v3
         with:
-          path: ~/.cache/rebar3/hex/hexpm/packages
+          path: _build
           key: lint-hex-${{ hashFiles(format('{0}{1}', github.workspace, '/rebar.lock')) }}
           restore-keys: lint-hex
-      - name: Lint
-        run: rebar3 lint
+      - name: Cache Dialyzer PLTs
+          uses: actions/cache@v3
```
zmstone: bad indentation?

.github/workflows/build.yml
```diff
@@ -100,4 +96,4 @@ jobs:
         uses: actions/upload-artifact@v1
         with:
           name: edoc
-          path: doc
+          path: doc
```
zmstone: Nit: don’t remove the last linebreak

spencerdcarlson:
I should have read [this](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) first. 🤦‍♂️. This should be correct now. 

Each `_build` cache key includes the OTP version at a minimum. Running the same pipeline multiple times will always result in a cache hit for each job. We may get some upstream, cross branch cache hits. A cache hit looks like it shaves off about 2 min.

spencerdcarlson:
I'm going to try and dig into the flapping tests a little. So far this is what I have observed:

* It seems to be the `t_restart_previous_*` test cases.
Example: `t_restart_previous_producers`
* Running a specific flapping test in isolation multiple times does not reproduce the issue.
  ```bash
  rebar3 ct --suite=test/brod_client_SUITE --case t_restart_previous_producers --repeat 100
  All 1 tests passed. (100 times)
  ```
* Running the suite that contains the flapping test can occasionally reproduce the issue.
  ```bash
  rebar3 ct --suite=test/brod_client_SUITE --repeat 10
  Failed 1 tests. Passed 13 tests. (2 times)
  All 14 tests passed. (8 times)
  ===> Failures occurred running tests: 2
  ```

**Stacktrace**
```bash
{{badmatch,{error,{producer_not_found,<<"brod-client-SUITE-topic">>}}},
 [{brod_client_SUITE,t_restart_previous_producers,1,
                     [{file,"/Users/sdc/code/sdc/brod/test/brod_client_SUITE.erl"},
                      {line,255}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1783}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1292}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,1224}]}]}
```

It does not seem like this MR is what is causing the tests to flap — This appears to be a preexisting issue. 

It _seems_ like the race condition has to do with shutting down the producer or consumer. 
* If I add a half second sleep after the call to `brod_client:stop_producer/2` ([L253-L254](https://github.com/kafka4beam/brod/blob/master/test/brod_client_SUITE.erl#L253-L254)) and `brod_client:stop_consumer/2` ([L269-L270](https://github.com/kafka4beam/brod/blob/master/test/brod_client_SUITE.erl#L269-L270)) I can run the SUITE 100 times with no flapping errors.
* When the test flaps the call to `brod:start_producer/3` or `brod:start_consumer/3` actually falls into the ` %% already started` clause ([producer](https://github.com/kafka4beam/brod/blob/master/src/brod_client.erl#L168), [consumer](https://github.com/kafka4beam/brod/blob/master/src/brod_client.erl#L188)), meaning that the even though stop was just called, the process is currently alive.

It looks like tests are being run synchronously and ordered since there are no groups. There is only 1 call to stop producer and consumer in all tests, so I don't see how this could be contention between tests. 

The stop code also looks like it is synchronous eventually executing `supervisor3:terminate_child/2` and `supervisor3:delete_child/2`I don't know how it would be possible for `brod_client:stop_producer/2` to return and the process not be fully shutdown before the next line in the test case executes ` brod:start_producer/3` but that seems like what is happening. Start is getting called while the pid is still terminating so we see "already started" and then it shuts down and `brod:get_producer/3` fails. 

Moving this to https://github.com/kafka4beam/brod/issues/504

spencerdcarlson:
Rebased and squashed. Should be finalized.

test/brod_topic_subscriber_SUITE.erl
```diff
@@ -388,7 +388,7 @@ check_state_continuity(Trace) ->
 
 %% Check that for any `init' there is a `terminate':
 check_init_terminate(Trace) ->
-  ?strict_causality( #{worker_id := _ID, ?snk_kind := topic_subscriber_init}
+  ?strict_causality( #{worker_id := _WorkerID, ?snk_kind := topic_subscriber_init}
```
zmstone: Was the intention to assert same ID in the two traces?

spencerdcarlson: Reverted. At some point (I think on another branch) I was getting a lint error. I wasn't able to reproduce it, so I just reverted it.

PR #502, opened 2022-05-11
Added hut, fixed compile warnings for test

spencerdcarlson:
* Updated snabbkaffe to `1.0.1`
* Added hut as an explicit test dependency 
* Removed test compile warnings

### DIFF
--- a/test/brod_topic_subscriber_SUITE.erl
+++ b/test/brod_topic_subscriber_SUITE.erl
@@ -373,8 +373,10 @@ check_received_messages(Expected, Trace) ->
 
 %% Check that state of callback module is correctly passed around
 %% between the calls:
-check_state_continuity(_WorkerId, Trace) ->
-  snabbkaffe:strictly_increasing([S || #{ state := S} <- Trace]).
+check_state_continuity(WorkerId, Trace) ->
+  snabbkaffe:strictly_increasing(
+    lists:filtermap(fun(#{worker_id := W, state := S}) when W =:= WorkerId -> {true, S};
+                       (_) -> false end, Trace)).
 
 check_state_continuity(Trace) ->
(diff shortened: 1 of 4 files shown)


test/brod_topic_subscriber_SUITE.erl
```diff
@@ -373,10 +373,8 @@ check_received_messages(Expected, Trace) ->
 
 %% Check that state of callback module is correctly passed around
 %% between the calls:
-check_state_continuity(WorkerId, Trace) ->
-  snabbkaffe:strictly_increasing([S || #{ worker_id := WorkerId
-                                        , state := S
-                                        } <- Trace]).
+check_state_continuity(_WorkerId, Trace) ->
+  snabbkaffe:strictly_increasing([S || #{ state := S} <- Trace]).
```
spencerdcarlson: I'm not sure that that is the desired outcome here. This change fixes the warnings but it seems suspicious to not be using `WorkerId` but the comprehension is not returning it. Should it be something alone these lines?

```erlang
[{S, WorkerId} || #{ worker_id := WorkerId, state := S} <- Trace]
```

spencerdcarlson:
What are your thoughts about setting `warnings_as_errors` in `erl_opts`?

test/brod_topic_subscriber_SUITE.erl
zmstone: @ieQu1 is this intended to be a filter ?
because as @spencerdcarlson noted, the `WorkerId` is shadowed.

maybe should be
```
lists:filtermap(fun(#{worker_id := W, state := S}) when W =:= WorkerId -> {true, S};
                   (_) -> false end, Trace)
```

zmstone (approved):

Issue #504, opened 2022-05-13
Flapping Common Tests


spencerdcarlson:
* It seems to be the `t_restart_previous_*` test cases.

Example: `t_restart_previous_producers`
* Running a specific flapping test in isolation multiple times does not reproduce the issue.
  ```bash
  rebar3 ct --suite=test/brod_client_SUITE --case t_restart_previous_producers --repeat 100
  All 1 tests passed. (100 times)
  ```
* Running the suite that contains the flapping test can occasionally reproduce the issue.
  ```bash
  rebar3 ct --suite=test/brod_client_SUITE --repeat 10
  Failed 1 tests. Passed 13 tests. (2 times)
  All 14 tests passed. (8 times)
  ===> Failures occurred running tests: 2
  ```

**Stacktrace**
```bash
{{badmatch,{error,{producer_not_found,<<"brod-client-SUITE-topic">>}}},
 [{brod_client_SUITE,t_restart_previous_producers,1,
                     [{file,"/Users/sdc/code/sdc/brod/test/brod_client_SUITE.erl"},
                      {line,255}]},
  {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1783}]},
  {test_server,run_test_case_eval1,6,[{file,"test_server.erl"},{line,1292}]},
  {test_server,run_test_case_eval,9,[{file,"test_server.erl"},{line,1224}]}]}
```

spencerdcarlson:
This _seems_ like the race condition between shutting down the producer or consumer and the subsequent call to start it.

* If I add a half second sleep after the call to `brod_client:stop_producer/2` ([L253-L254](https://github.com/kafka4beam/brod/blob/master/test/brod_client_SUITE.erl#L253-L254)) and `brod_client:stop_consumer/2` ([L269-L270](https://github.com/kafka4beam/brod/blob/master/test/brod_client_SUITE.erl#L269-L270)) I can run the SUITE 100 times with no flapping errors.
  ```erlang
  t_restart_previous_producers(Config) when is_list(Config) ->
    Client = t_restart_previous_producers,
    Topic = ?TOPIC,
    ok = start_client(?HOSTS, Client),
    ok = brod:start_producer(Client, Topic, []),
    ok = brod_client:stop_producer(Client, Topic),
    timer:sleep(500), %% Not a valid solution, just identifying the issue.
    ?assertEqual(ok, brod:start_producer(Client, Topic, [])),
[...]

PR #505, opened 2022-05-19
Ets race condition

spencerdcarlson:
Attempt at addressing https://github.com/kafka4beam/brod/issues/504

### DIFF
--- a/src/brod_client.erl
+++ b/src/brod_client.erl
@@ -145,8 +145,6 @@ get_producer(Client, Topic, Partition) ->
     {ok, Pid} ->
       {ok, Pid};
     {error, {producer_not_found, Topic}} = Error ->
-      %% try to start a producer for the given topic if
-      %% auto_start_producers option is enabled for the client
       maybe_start_producer(Client, Topic, Partition, Error);
     Error ->
       Error
@@ -420,19 +418,37 @@ shutdown_pid(_) ->
         {ok, pid()} | {error, get_worker_error()}.
 get_partition_worker(ClientPid, Key) when is_pid(ClientPid) ->
   case erlang:process_info(ClientPid, registered_name) of
-    {registered_name, ClientId} when is_atom(ClientId) ->
-      get_partition_worker(ClientId, Key);
+    {registered_name, ClientId} -> get_partition_worker(ClientId, Key);
+
     _ ->
       %% This is a client process started without registered name
       %% have to call the process to get the producer/consumer worker
       %% process registration table.
-      case safe_gen_call(ClientPid, get_workers_table, infinity) of
-        {ok, Ets}       -> lookup_partition_worker(ClientPid, Ets, Key);
-        {error, Reason} -> {error, Reason}
-      end
+      get_partition_worker_with_ets(ClientPid, Key)
   end;
+
 get_partition_worker(ClientId, Key) when is_atom(ClientId) ->
-  lookup_partition_worker(ClientId, ?ETS(ClientId), Key).
+  case lookup_partition_worker(ClientId, ?ETS(ClientId), Key) of
+    {ok, Pid} ->
+      %% This is an attempt to help with a race condition
+      %% between ets lookups and what is being supervised.
+      %% If the worker process is returned form ets,
+      %% but it is not alive then there must be
+      %% an inflight worker deregistration request.
+      case is_process_alive(Pid) of
+        true -> {ok, Pid};
+        false -> get_partition_worker_with_ets(ClientId, Key)
+      end;
+    Other -> Other
+  end.
+
+-spec get_partition_worker_with_ets(client(), partition_worker_key()) ->
+  {ok, pid()} | {error, get_worker_error()}.
+get_partition_worker_with_ets(Client, Key) ->
+  case safe_gen_call(Client, get_workers_table, infinity) of
+    {ok, Ets} -> lookup_partition_worker(Client, Ets, Key);
+    {error, Reason} -> {error, Reason}
+  end.
 
 -spec lookup_partition_worker(client(), ets:tab(), partition_worker_key()) ->
         {ok, pid()} | { error, get_worker_error()}.
@@ -783,6 +799,8 @@ do_get_partitions_count(TopicMetadata) ->
                            partition(), {error, any()}) ->
         ok | {error, any()}.
 maybe_start_producer(Client, Topic, Partition, Error) ->
+  %% try to start a producer for the given topic if
+  %% auto_start_producers option is enabled for the client
   case safe_gen_call(Client, {auto_start_producer, Topic}, infinity) of
     ok ->
       get_partition_worker(Client, ?PRODUCER_KEY(Topic, Partition));
--- a/test/brod_client_SUITE.erl
+++ b/test/brod_client_SUITE.erl
@@ -399,8 +399,10 @@ retry_writer_loop(Parent, ProduceFun, LastResult) ->
   Result = ProduceFun(),
   %% assert result patterns
   case Result of
-    ok              -> ok;
-    {error, Reason} -> ?assertMatch({producer_down, _}, Reason)
+    ok -> ok;
+    {error, client_down} -> ok;
+    {error, {producer_down, _}} -> ok;
+    {error, Reason} -> error({unexpected_error, Reason})
   end,
   %% tell parent about produce return value changes
   case Result =/= LastResult of
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,6 @@
 * 3.16.3
   * Fix specs for `delete_topics`.
+  * Ensure that partition worker process is alive before returning it.
 * 3.16.2
   * Update kafka_protocol from 4.0.1 to 4.0.3.
     Prior to this change the actual time spent in establishing a


spencerdcarlson:
It probably makes more sense to refactor [get_partition_worker/2, lookup_partition_worker/3](https://github.com/kafka4beam/brod/blob/master/src/brod_client.erl#L419-L455) to always require a call before hitting ets, as opposed to how it currently only does it when the client does not have a registered name.

src/brod_client.erl
```diff
@@ -418,21 +419,23 @@ shutdown_pid(_) ->
...
-        {ok, Ets}       -> lookup_partition_worker(ClientPid, Ets, Key);
-        {error, Reason} -> {error, Reason}
-      end
-  end;
-get_partition_worker(ClientId, Key) when is_atom(ClientId) ->
-  lookup_partition_worker(ClientId, ?ETS(ClientId), Key).
+get_partition_worker(Client, Key) ->
+  %% Added potentially gratuitous call to `get_workers_table` to flush Client inbox
+  %% which ensures that ets is in sync with the children being supervised.
+  case {is_atom(Client), safe_gen_call(Client, get_workers_table, infinity)}  of
```
zmstone: This is a very hot code path for producers,
the unconditional gen call is likely a performance killer.

Maybe we can make gen call if:
Lookup failed, or lookup returned a process which is `not erlang:is_process_alive(Pid)`
So the overhead added is, in most of the cases, just `is_process_alive`

A side note: all “worker discovery” have to face the race condition of: work process being down after discovery when submitting a task to it.
This change only lowers the chance of such race.

src/brod_client.erl
```diff
@@ -418,21 +418,39 @@ shutdown_pid(_) ->
 
 -spec get_partition_worker(client(), partition_worker_key()) ->
         {ok, pid()} | {error, get_worker_error()}.
-get_partition_worker(ClientPid, Key) when is_pid(ClientPid) ->
-  case erlang:process_info(ClientPid, registered_name) of
-    {registered_name, ClientId} when is_atom(ClientId) ->
-      get_partition_worker(ClientId, Key);
+get_partition_worker(Client, Key) when is_atom(Client) ->
+  case erlang:whereis(Client) of
+    ClientPid when is_pid(ClientPid) ->
```
spencerdcarlson: Not sure if you have a preference here. We could call `get_partition_worker/2` and add a function override match on `undefined` or `_` now that we have a pid or know that it's an invalid pid. If we have a valid pid, then we would immediately be looking up its registered name. 😆

src/brod_client.erl
```diff
@@ -418,21 +418,39 @@ shutdown_pid(_) ->
...
+          %% This is an attempt to help with a race condition
+          %% between ets lookups and what is being supervised.
+          %% Adding a gratuitous call to `get_workers_table`
+          %% to flush the Client's inbox of any inflight worker registration or deregistration.
+          case safe_gen_call(ClientPid, get_workers_table, infinity) of
+            {ok, _} -> lookup_partition_worker(Client, ?ETS(Client), Key);
+            {error, Reason} -> {error, Reason}
+          end
+      end;
+    _ -> {error, client_down}
```
spencerdcarlson: This keeps the status quo. I figure there is no need to call  `get_workers_table` if we know its an invalid pid. We could add a more descriptive return `{error, invalid_client}` or just keep it. Thoughts?

src/brod_client.erl
spencerdcarlson: 100%, My understanding is that just because we "flush" the inbox right before we call ETS doesn't mean that something else couldn't sneak in a cast right before we actually read from ETS.

spencerdcarlson: Curious if you have any insight into the note about [supervisor:which_children/1](https://www.erlang.org/doc/man/supervisor.html#which_children-1)
>Notice that calling this function when supervising many children under low memory conditions can cause an out of memory exception.

It seems like being able to use that function would solve so many headaches in so many scenarios. But because of the vagueness around _many_ and _low_ in the above statement it appears to be off limits.

src/brod_client.erl
zmstone: Better not to introduce a new reason code in a patch fix.

src/brod_client.erl
```diff
@@ -418,21 +418,41 @@ shutdown_pid(_) ->
...
 -spec get_partition_worker(client(), partition_worker_key()) ->
         {ok, pid()} | {error, get_worker_error()}.
-get_partition_worker(ClientPid, Key) when is_pid(ClientPid) ->
-  case erlang:process_info(ClientPid, registered_name) of
-    {registered_name, ClientId} when is_atom(ClientId) ->
-      get_partition_worker(ClientId, Key);
+get_partition_worker(Client, Key) when is_atom(Client) ->
+  case erlang:whereis(Client) of
+    ClientPid when is_pid(ClientPid) ->
+      case erlang:is_process_alive(ClientPid) of
```
zmstone: where_is returning a pid is an indication of the process being alive.
There should be no need to check again immediately.

src/brod_client.erl
```diff
@@ -418,21 +418,41 @@ shutdown_pid(_) ->
...
+        true -> lookup_partition_worker(Client, ?ETS(Client), Key);
+
+        _ ->
+          %% This is an attempt to help with a race condition
+          %% between ets lookups and what is being supervised.
+          %% Adding a gratuitous call to `get_workers_table`
+          %% to flush the Client's inbox of any inflight worker registration or deregistration.
+          get_partition_worker_with_ets(ClientPid, Key)
```
zmstone: Making a gen call to a dead process will always return error.

src/brod_client.erl
```diff
@@ -418,21 +418,41 @@ shutdown_pid(_) ->
...
+      end;
+    _ -> {error, client_down}
+  end;
+
+get_partition_worker(ClientPid, Key) when is_pid(ClientPid)->
+  case {erlang:is_process_alive(ClientPid), erlang:process_info(ClientPid, registered_name)} of
```
zmstone: No need to check liveness together with process_info, because process_info always return undefined for dead process.

src/brod_client.erl
```diff
@@ -418,21 +418,41 @@ shutdown_pid(_) ->
...
+    {true, {registered_name, ClientName}} when is_atom(ClientName) ->
```
zmstone: Could registered name be something other than an atom? I.e. the is_atom check is not needed?

src/brod_client.erl
```diff
@@ -418,21 +418,41 @@ shutdown_pid(_) ->
...
+      lookup_partition_worker(ClientName, ?ETS(ClientName), Key);
```
zmstone:
```suggestion
      case lookup_partition_worker(ClientName, ?ETS(ClientName), Key) of
        {ok, Pid} ->
          case is_process_alive(Pid) of
            true -> {ok, Pid};
            false -> get_partition_worker_with_ets(ClientPid, Key)
          end
        Other ->
          Other
      end.
```

src/brod_client.erl
```diff
@@ -420,19 +420,36 @@ shutdown_pid(_) ->
...
+  case lookup_partition_worker(ClientId, ?ETS(ClientId), Key) of
+    {ok, Pid} ->
+      %% This is an attempt to help with a race condition
+      %% between ets lookups and what is being supervised.
+      %% If the worker process is returned form ets,
+      %% but it is not alive then there must be
+      %% an inflight worker deregistration request.
+      case is_process_alive(Pid) of
+        true -> {ok, Pid};
+        false -> partition_worker_error(Key)
```
spencerdcarlson: ~~If I call `get_workers_table` then by that time the actual client has terminated.~~
This is because I was passing in the pid of the partition worker to try and lookup the `get_workers_table` when it needs to be the cllient atom or pid 🤦  — [comment](https://github.com/kafka4beam/brod/pull/505#discussion_r880920021)

zmstone: It is true when client is shutting down.
But it could also be that just one producer worker is restarting?
So I think it’s better to call get_workers_table
on gen call `exit` exception: return client_down
Otherwise return the (possibly new) pid from ets lookup.

src/brod_client.erl
```diff
@@ -144,9 +144,9 @@ get_producer(Client, Topic, Partition) ->
   case get_partition_worker(Client, ?PRODUCER_KEY(Topic, Partition)) of
     {ok, Pid} ->
       {ok, Pid};
+    {error, client_down} = Error ->
+      maybe_start_producer(Client, Topic, Partition, Error);
```
spencerdcarlson: Do we need to restart the producer regardless of if `auto_start_producers` is enabled?

zmstone: if client is down, the maybe_start_producer call is unlikely to succeed ?

zmstone: if client is down, the maybe_start_producer call is unlikely to succeed ?

src/brod_client.erl
```diff
@@ -186,6 +186,9 @@ start_consumer(Client, TopicName, ConsumerConfig) ->
   case get_consumer(Client, TopicName, _Partition = 0) of
     {ok, _Pid} ->
       ok; %% already started
+    {error, client_down} ->
+      Call = {start_consumer, TopicName, ConsumerConfig},
+      safe_gen_call(Client, Call, infinity);
```
zmstone: same here, if client is down, we'll need to ensure client is started, then start consumer again.

zmstone: same here, if client is down, we'll need to ensure client is started, then start consumer again.

test/brod_client_SUITE.erl
```diff
@@ -400,6 +400,7 @@ retry_writer_loop(Parent, ProduceFun, LastResult) ->
   %% assert result patterns
   case Result of
     ok              -> ok;
+    {error, client_down} -> ok;
     {error, Reason} -> ?assertMatch({producer_down, _}, Reason)
```
zmstone:
```suggestion
  case Result of
    ok -> ok;
    {error, client_down} -> ok;
    {error, {producer_down, _}} -> ok;
    {error, Reason} -> error({unexpected_error, Reason})
```

src/brod_client.erl
```diff
@@ -420,19 +423,37 @@ shutdown_pid(_) ->
...
+  case lookup_partition_worker(ClientId, ?ETS(ClientId), Key) of
+    {ok, Pid} ->
+      %% This is an attempt to help with a race condition
+      %% between ets lookups and what is being supervised.
+      %% If the worker process is returned form ets,
+      %% but it is not alive then there must be
+      %% an inflight worker deregistration request.
+      case is_process_alive(Pid) of
+        true -> {ok, Pid};
+        false -> get_partition_worker_with_ets(Pid, Key)
```
spencerdcarlson: Oooops - `get_partition_worker_with_ets(ClientId, Key)`

src/brod_client.erl
spencerdcarlson: We can remove all of these. I put them in because of this [typo](https://github.com/kafka4beam/brod/pull/505/files#r880920021) which caused a return type of `{error, client_down}` instead of `{error,{[producer|consumer]_not_found,<<"brod-client-SUITE-topic">>}}`.

spencerdcarlson:
I can run `rebar3 ct --suite=test/brod_client_SUITE --repeat 100` with no flapping tests 🥳

zmstone (approved):

zmstone:
Great fix @spencerdcarlson .
forgot one last thing: please update changelog.md

PR #506, opened 2022-05-26
Proposal: Optimize tests

spencerdcarlson:
Submitting this MR as an example. What are your thoughts on: 
* leveraging groups to run tests in parallel 
* Making tests repeatable (where it makes sense)
* Making tests explicitly dependent (where it makes sense)
* Adding ct:logs

### DIFF
--- a/test/brod_SUITE.erl
+++ b/test/brod_SUITE.erl
@@ -26,8 +26,7 @@
         ]).
 
 %% Test cases
--export([ t_create_topics/0
-        , t_create_topics/1
+-export([ t_create_topics/1
         , t_delete_topics/1
         , t_delete_topics_not_found/1
         ]).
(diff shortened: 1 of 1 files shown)


spencerdcarlson:
In the included example, I would prefer to combine the create and delete into one test that first creates and then deletes the same topic instead of passing state between two tests. That would allow for true isolation and parallelism. I did it this way here as an example of what might be the better option for some of the other test modules. 

I'm curious if you would welcome me doing some test refactoring to try and make them more explicit, isolated, repeatable, and potentially faster overall execution time.

zmstone:
some of the test cases (integration test with real Kafka) may use the same topic-partitions, e.g. `test-topic` to send / receive messages.
otherwise running in parallel should work ok.

> I'm curious if you would welcome me doing some test refactoring to try and make them more explicit, isolated, repeatable, and potentially faster overall execution time.

Sure, this is very much welcome.

PR #512, opened 2022-06-18
WIP: add telemetry

spencerdcarlson:


rebar.config
```diff
@@ -30,8 +31,9 @@
            , {meck, "0.9.2"}
            , {proper, "1.4.0"}
            , {snabbkaffe, "1.0.1"}
+           , {telemetry, "~> 1.0"}
     ]},
-    {erl_opts, [warnings_as_errors, {d, build_brod_cli}]}
+    {erl_opts, [warnings_as_errors, {d, build_brod_cli}, {d, brod_use_telemetry}]}
```
zmstone, line 36: I'm convinced this is not necessary.
Please remove it.

zmstone:
Thanks for the PR.

BTW. I originally wanted a solution without `{telemetry, "~> 1.0"}` in the dependency (and the `brod_use_telemetry` to be **undefined** by default).
So to push the responsibility to the wrapping project.

zmstone:
I have changed the base to `dev/telemetry`
let's start instrumenting the code, preferably increment with small PRs.

maxno-kivra:
This would be very nice to have. What is needed to get this in?

akosasante:
:wave:  Also looking to help push this forward, @zmstone , how can I help?

mikpe:
Use dependency injection.
Have measurement points invoke a local, "virtual", telemetry module.
That module checks if a telemetry callback module has been registered in the environment: if it has, pass the telemetry on to it, if not, do nothing.
The end result is that this project has no hard-coded dependencies in any telemetry framework, but the user can hook one up via the callback module _they_ control.

PR #515, opened 2022-06-30
Fix `get_committed_offset` return specs

SerikDM:
Fix specs according to expected format in: https://github.com/kafka4beam/brod/blob/f88130e051fc2e317e744217f9e05071378d4629/src/brod_group_subscriber_v2.erl#L281

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,5 @@
+* 3.16.5
+  * Fix specs for `get_committed_offset`
 * 3.16.4
   * Update kafka_protocol from 4.0.3 to 4.1.0. kafka_protocol 4.1.0 support a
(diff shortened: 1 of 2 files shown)


zmstone:
thanks could you help to update Changelog.md ?

SerikDM:
> thanks could you help to update Changelog.md ?

Sure, added a line about fixing specs

changelog.md
```diff
@@ -1,3 +1,5 @@
+* 3.16.5
+  * Fix specs for `get_committed_offset`
```
zmstone: Please add module name too.

SerikDM: Added

PR #522, opened 2022-08-05
Fix the Elixir sample code and update all dependencies

robsonpeixoto:


contrib/examples/elixir/lib/brod_sample/application.ex
```diff
@@ -7,8 +7,11 @@ defmodule BrodSample.Application do
 
   def start(_type, _args) do
     children = [
-      BrodSample.GroupSubscriber
+      BrodSample.GroupSubscriberV2
+      # FIXME: If start both process only the "BrodSample.GroupSubscriberV2" consumes messages.
```
robsonpeixoto: Any ideia of how to solve it?

robsonpeixoto: It solved my problem: https://github.com/kafka4beam/brod/pull/522/commits/3f92dc69d7e0129fc2737cbfbdc02523bc5f50f6

contrib/examples/elixir/lib/brod_sample/group_subscriber.ex
```diff
@@ -36,13 +37,14 @@ defmodule BrodSample.GroupSubscriber do
   end
 
   def handle_message(
```
GabriellaDiasA, line 39:
```suggestion

  @impl true
  def handle_message(
```

GabriellaDiasA (commented):
I think `init/2` is also a callback, should probably add the `@impl true` line there as well

PR #533, opened 2022-09-21
call get_partitions_count with checking allow_topic_auto_creation option

dbegunkov:


dbegunkov:
The fix solveds problem with autocreation topics when we call method brod:get_partitions_count/2, if the flag of config allow_topic_auto_creation = false. For example, when we using kaffe lib (https://github.com/spreedly/kaffe), we call method `Kaffe.Producer. produce /3`, that called method :brod.brod:get_partitions_count/2, to calculate needed partition.

zmstone:
Hi @dbegunkov 
Thank you for the PR.
I have it covered here:  https://github.com/kafka4beam/brod/pull/541
Made `brod_client` to cache the the `unknown_top_partition` result.

dbegunkov:
Thank for your fix this problem!

zmstone:
released in `3.16.5`

PR #534, opened 2022-10-07
Improve documentation

xxdavid:
Hi, thank you for making `brod`! We use it in our company and we really felt the absence of documentation for some of the aspects so I spent a few days trying to improve it.

The diff is quite long but I've tried to split it into meaningful commits. I've commented on some decisions I made into the commit messages but feel free to ask if there is anything unclear.

[edit by @zmstone ]:
Things to be done before merge:

- [x] resolve revew comments
- [x] squash commits
- [x] update changelog.md (3.16.6)

### DIFF
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -817,8 +817,7 @@ start_link_topic_subscriber(Config) ->
   brod_topic_subscriber:start_link(Config).
 
 %% @equiv create_topics(Hosts, TopicsConfigs, RequestConfigs, [])
--spec create_topics([endpoint()], [topic_config()], #{timeout => kpro:int32(),
-                    validate_only => boolean()}) ->
+-spec create_topics([endpoint()], [topic_config()], #{timeout => kpro:int32()}) ->
         ok | {error, any()}.
 create_topics(Hosts, TopicConfigs, RequestConfigs) ->
   brod_utils:create_topics(Hosts, TopicConfigs, RequestConfigs).
@@ -827,8 +826,8 @@ create_topics(Hosts, TopicConfigs, RequestConfigs) ->
 %%
 %% Return the message body of `create_topics', response.
 %% See `kpro_schema.erl' for struct details.
--spec create_topics([endpoint()], [topic_config()], #{timeout => kpro:int32(),
-                    validate_only => boolean()}, conn_config()) ->
+-spec create_topics([endpoint()], [topic_config()], #{timeout => kpro:int32()},
+                    conn_config()) ->
         ok | {error, any()}.
 create_topics(Hosts, TopicConfigs, RequestConfigs, Options) ->
   brod_utils:create_topics(Hosts, TopicConfigs, RequestConfigs, Options).
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -819,7 +819,7 @@ start_link_topic_subscriber(Config) ->
 %% @equiv create_topics(Hosts, TopicsConfigs, RequestConfigs, [])
 -spec create_topics([endpoint()], [topic_config()], #{timeout => kpro:int32(),
                     validate_only => boolean()}) ->
-        ok | {ok, kpro:struct()} | {error, any()}.
+        ok | {error, any()}.
 create_topics(Hosts, TopicConfigs, RequestConfigs) ->
   brod_utils:create_topics(Hosts, TopicConfigs, RequestConfigs).
 
@@ -829,7 +829,7 @@ create_topics(Hosts, TopicConfigs, RequestConfigs) ->
 %% See `kpro_schema.erl' for struct details.
 -spec create_topics([endpoint()], [topic_config()], #{timeout => kpro:int32(),
                     validate_only => boolean()}, conn_config()) ->
-        ok | {ok, kpro:struct()} | {error, any()}.
+        ok | {error, any()}.
 create_topics(Hosts, TopicConfigs, RequestConfigs, Options) ->
   brod_utils:create_topics(Hosts, TopicConfigs, RequestConfigs, Options).
 
--- a/src/brod.erl
+++ b/src/brod.erl
@@ -133,8 +133,6 @@
              , connection/0
              , conn_config/0
              , consumer_config/0
-             , consumer_option/0
-             , consumer_options/0
              , endpoint/0
              , error_code/0
              , fetch_opts/0
@@ -206,8 +204,12 @@
 -type offset_time() :: offset()
                      | ?OFFSET_EARLIEST
                      | ?OFFSET_LATEST.
--type message() :: kpro:message().
+-type message() :: kpro:message(). %% A record with offset, key, value, ts_type, ts, and headers.
 -type message_set() :: #kafka_message_set{}.
+%% A record with topic, partition, high_wm_offset (max offset of the partition), and messages.
+%%
+%% See <a href="https://github.com/kafka4beam/brod/blob/master/include/brod.hrl#L26">
+%% the definition</a> for more information.
 -type error_code() :: kpro:error_code().
 
 %% producers
@@ -224,17 +226,20 @@
 
 
 %% consumers
--type consumer_option() :: begin_offset
-                         | min_bytes
-                         | max_bytes
-                         | max_wait_time
-                         | sleep_timeout
-                         | prefetch_count
-                         | prefetch_bytes
-                         | offset_reset_policy
-                         | size_stat_window.
--type consumer_options() :: [{consumer_option(), integer()}].
--type consumer_config() :: brod_consumer:config().
+-type consumer_config() :: [ {begin_offset,        offset_time()}
+                           | {min_bytes,           non_neg_integer()}
+                           | {max_bytes,           non_neg_integer()}
+                           | {max_wait_time,       integer()}
+                           | {sleep_timeout,       integer()}
+                           | {prefetch_count,      integer()}
+                           | {prefetch_bytes,      non_neg_integer()}
+                           | {offset_reset_policy, brod_consumer:offset_reset_policy()}
+                           | {size_stat_window,    non_neg_integer()}
+                           | {isolation_level,     brod_consumer:isolation_level()}
+                           ].
+%% Consumer configuration.
+%%
+%% The meaning of the options is documented at {@link brod_consumer:start_link/5}.
 -type connection() :: kpro:connection().
 -type conn_config() :: [{atom(), term()}] | kpro:conn_config().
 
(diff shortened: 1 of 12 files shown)


xxdavid:
Could you please take a look on it? 🙂 @zmstone

xxdavid:
Thank you for the approval of pipelines. I replaced the URL (in 898dad5) on the long line a with a shortened URL (https://tinyurl.com/kafka-fetch-api-doc) so I think the failed workflow should pass now.

(I am also not really sure whether to include the link at all, as Kafka is moving from this docs on Confluence (if I understand it correctly) but the [new documentation](https://kafka.apache.org/protocol.html) is much terser and lacks some useful info.)

zmstone:
> new documentation

maybe this ? https://kafka.apache.org/protocol.html#The_Messages_Fetch

src/brod.erl
```diff
@@ -427,23 +442,52 @@ stop_client(Client) when is_atom(Client) ->
...
                         ok | {error, any()}.
 start_producer(Client, TopicName, ProducerConfig) ->
   brod_client:start_producer(Client, TopicName, ProducerConfig).
 
-%% @doc Dynamically start a topic consumer.
-%% @see brod_consumer:start_link/5. for details about consumer config.
+%% @doc Dynamically start topic consumer(s) and register it in the client.
+%%
+%% A {@link brod_consumer} is started for each partition of the given topic.
+%% Note that you can have only one consumer per client-topic-partition.
```
zmstone:
```suggestion
%% Note that you can have only one consumer per client-topic.
```

src/brod.erl
```diff
@@ -627,12 +720,24 @@ produce_sync_offset(Client, Topic, Partition, Key, Value) ->
...
+%%
+%% This way, you can turn asynchronous requests, made by {@link produce/5}
+%% and friends, into synchronous ones.
+%%
+%% Example:
+%% ```
+%% {ok, CallRef} = brod:produce(
+%%   brod_client_1, <<"my_topic">>, 0, <<"some-key">>, <<"some-value">>)
+%% ). % returns immediately
+%% ok = brod:sync_produce_request(CallRef). % waits and returns after the ack is received
```
zmstone:
```suggestion
%% brod:sync_produce_request(CallRef, 5_000). % waits and returns after the ack is received or timed out
```

src/brod.erl
```diff
@@ -711,6 +841,16 @@ unsubscribe(ConsumerPid) ->
 unsubscribe(ConsumerPid, SubscriberPid) ->
   brod_consumer:unsubscribe(ConsumerPid, SubscriberPid).
 
+%% @doc Acknowledge that a message has been processed.
```
zmstone:
```suggestion
%% @doc Acknowledge that one or more messages have been processed.
%%
%% {@link brod_consumer} sends message-sets to the subscriber process, and keep
%% the messages in a 'pending' queue.
%% The subscriber may choose to ack any received offset.
%% Acknowledging a greater offset will automatically acknowledge
%% the messages before this offset.
%% For example, if message `[1, 2, 3, 4]` have been sent to (as one or more message-sets)
%% to the subscriber, the subscriber may acknowledge with offset `3` to indicate that
%% the first three messages are successfully processed, leaving behind only message `4`
%% pending.
%%
```

src/brod.erl
```diff
@@ -711,6 +841,16 @@ unsubscribe(ConsumerPid) ->
...
+%%
+%% To provide a mechanism to handle backpressure, brod requires all messages
+%% sent to a subscriber to be acked when they are processed. If there are too
+%% many not-acked messages received by the subscriber, the consumer will stop
+%% fetching new ones so the subscriber won't get overwhelmed.
```
zmstone:
```suggestion
%% The 'pending' queue has a size limit (see `prefetch_count` consumer config)
%% which is to provide a mechanism to handle back-pressure.
%% If there are too many messages pending on ack, the consumer will stop
%% fetching new ones so the subscriber won't get overwhelmed.
```

src/brod.erl
```diff
@@ -798,18 +940,64 @@ start_link_topic_subscriber(Config) ->
...
+%%    <li>`replication_factor'
+%%
+%%      The number of replicas to create for each partition in the topic,
+%%      or -1 if we are either specifying a manual partition assignment
+%%      or using the default replication factor.</li>
+%%
+%%    <li>`assignments'
+%%
+%%      The manual partition assignment, or the empty array if we are using
+%%      automatic assignment. It is a list of maps (or tuple lists) with the
```
zmstone:
```suggestion
%%      The manual partition assignment, or the empty list if we let Kafka
%%      automatically assign them. It is a list of maps (or tuple lists) with the
```

src/brod.erl
```diff
@@ -798,18 +940,64 @@ start_link_topic_subscriber(Config) ->
...
+%%      following keys: `partition_index' and `broker_ids' (a of brokers to
+%%      place the partition on).</li>
```
zmstone:
```suggestion
%%      following keys: `partition_index' and `broker_ids' (a list of of brokers to
%%      place the partition on).</li>
```

src/brod.erl
```diff
@@ -891,20 +1153,71 @@ fetch(ConnOrBootstrap, Topic, Partition, Offset) ->
...
+%% copied to all replicas, incremented by 1), whichever is lower. In essence, this
+%% is the offset up to which it was possible to read the messages at the time of
+%% fetching. This is similar to what {@link resolve_offset/6} with `latest'
+%% returns. You can use this information to determine how far from the end of the
+%% topic you currently are. Note that when you use this offset as the start offset
+%% for a subseuqent call, an empty list of messages will be returned (assuming the
+%% topic hasn't changed, e.g. no new message arrived). Only when you use an offset
+%% greater than this one, `{error, offset_out_of_range}' will be returned.
+%%
+%% Note also that the messages are fetched only up to the end of a topic segment
```
zmstone:
```suggestion
%% Note also that Kafka batches messages in a message set only up to the end of a topic segment
```

src/brod.erl
```diff
@@ -891,20 +1153,71 @@ fetch(ConnOrBootstrap, Topic, Partition, Offset) ->
...
+%% in which the first retrieved message is, so there may actually be more
+%% messages in the topic than were retrieved (even in a static topic).
```
zmstone: not sure what a 'static topic' is, no producers appending messages while fetching ?
```suggestion
%% messages behind the last fetch offset even if the fetch size is significantly less than `max_bytes` provided in `fetch_opts()`.
```

zmstone (commented):
not done yet.
for my own record: continue from `brod_client.erl`

src/brod_consumer.erl
```diff
@@ -13,6 +13,23 @@
...
 
+%% @doc Kafka consumers work in poll mode. In brod, `brod_consumer' is the poller,
+%% which is constantly asking for more data from the kafka node which is a leader
+%% for the given partition.
+%%
+%% By subscribing to `brod_consumer' a process should receive the polled message
+%% sets (not individual messages) into its mailbox. Shape of the message is
+%% documented at {@link brod:subscribe/5}.
+%%
+%% Every message processed by the subscriber has to be acked by calling
```
zmstone:
```suggestion
%% Messages processed by the subscriber has to be acked by calling
```

src/brod_consumer.erl
```diff
@@ -13,6 +13,23 @@
...
+%% {@link ack/2} (or {@link brod:consume_ack/4}) to notify the consumer
+%% that more message can be fetched and sent to the subscriber and the
```
zmstone:
```suggestion
%% that all messages before the acknowledged offsets are processed,
%% hence more messages can be fetched and sent to the subscriber and the
```

src/brod.erl
```diff
@@ -711,6 +841,16 @@ unsubscribe(ConsumerPid) ->
 unsubscribe(ConsumerPid, SubscriberPid) ->
   brod_consumer:unsubscribe(ConsumerPid, SubscriberPid).
 
+%% @doc Acknowledge that a message has been processed.
+%%
+%% To provide a mechanism to handle backpressure, brod requires all messages
+%% sent to a subscriber to be acked when they are processed. If there are too
+%% many not-acked messages received by the subscriber, the consumer will stop
+%% fetching new ones so the subscriber won't get overwhelmed.
+%%
```
zmstone, line 862:
```suggestion
%%
%% Note, there is no range check done for the acknowledging offset, meaning if offset `[M, N]`
%% are pending to be acknowledged, acknowledging with `Offset > N` will cause all offsets to be
%% removed from the pending queue, and acknowledging with `Offset < M` has no effect.
%%
```

zmstone:
This is awesome @xxdavid.
Thank you so much.

zmstone:
And sorry for the slow review. 
I've finished my first read through and added a few suggestions.

src/brod.erl
xxdavid: > not sure what a 'static topic' is, no producers appending messages while fetching ?

Yes, this was what I meant by *static topic*.

Thank you for making it more clear. I am not sure what you mean by *fetch size* though.

xxdavid:
Thank you very much for the review and also for your insightful additions! I've accepted your changes (apart from one where I need clarification) and fixed the quotation marks (EDoc uses the weird `code' syntax) with subsequent fixups.

I am still quite unsure about the documentation of parameters for `fetch`. The confluence docs have a nice description of what ReplicaId, MaxWaitTime and MinBytes exactly mean. What do you think about it? I also thought about directly copying the description from the confluence docs to the brod `fetch` docs, but this also doesn't seem quite right.

src/brod.erl
zmstone: Sorry, I meant 'fetched'.
```
messages behind the last fetched offset even if the fetched size is significantly less than `max_bytes` provided in `fetch_opts()`.
```

zmstone:
> Thank you very much for the review and also for your insightful additions! I've accepted your changes (apart from one where I need clarification) and fixed the quotation marks (EDoc uses the weird `code' syntax) with subsequent fixups.
> 
> I am still quite unsure about the documentation of parameters for `fetch`. The confluence docs have a nice description of what ReplicaId, MaxWaitTime and MinBytes exactly mean. What do you think about it? I also thought about directly copying the description from the confluence docs to the brod `fetch` docs, but this also doesn't seem quite right.


`fetch_opts` is a bit different from the documented fields,
Kafka allows fetching multiple topic-partitions,
`brod` only supports fetching single topic-partition, so the structure is more 'flat',
and some of the fields such as `replica_id` is not exposed from `brod` because it's only used by Kafka followers.

We can maybe document each opts, and also provide the Kafka doc for reference.
https://github.com/kafka4beam/kafka_protocol/blob/36d44c8e453102cbaf1586a134a45f2b0067b41c/src/kpro_req_lib.erl#L84-L90

src/brod.erl
xxdavid: Aha, it makes sense now. I'll use your suggestion and commit it locally.

xxdavid:
> fetch_opts is a bit different from the documented fields,
Kafka allows fetching multiple topic-partitions,
brod only supports fetching single topic-partition, so the structure is more 'flat',
and some of the fields such as replica_id is not exposed from brod because it's only used by Kafka followers.

Aha, didn't know that. Good catch.

> We can maybe document each opts, and also provide the Kafka doc for reference.

This sounds good! I'll try to document it alongside the `kpro_req_lib:fetch_opts()` type and add it to my kafka4beam/kafka_protocol#106 PR.

zmstone:
Hi @xxdavid 
Just merged https://github.com/kafka4beam/kafka_protocol/pull/106
I edited this PRs description, added 3 TODO checkboxes.
Could you help to fix them then this is good to merge.

xxdavid:
Hi @zmstone, great! All of the comments seem to be resvoled here so I just checked the box. I squashed the fixups, added a changelong entry, and force-pushed.

zmstone (approved):

xxdavid:
Thanks! :tada: Could you please also publish the version to Hex? The latest version there is 3.16.4.

PR #539, opened 2022-11-16
Make user_data overridable

dbalakavi:
Making user-data of a brod_group_subscriber_v2 member overridable comes in handy at the time of partition assignment. The assignment logic can be be based on the member metadata. For instance, in usecases where certain members get sticky assignments while the others get round-robin assignments.


dbalakavi:
For posterity & context: this PR was for test purposes only (to begin with), and intended against a fork of this repository.

PR #553, opened 2023-02-10
Handle new opaque types in gen_server on OTP 25

jesperes:
OTP 25 introduced new types in `gen_server`, but some of these are opaque, breaking existing code. I could not find a suitable backwards compatible way of doing this without resorting to a kludge like this.


jesperes:
(Of course, I could have removed the type for `pending_sync`, but that didn't seem right either.)

mikpe (approved):

PR #556, opened 2023-03-15
Copy supervisor3 from klarna/supervisor3 into brod

gonzalobf:
I suppose this will be a little bit controversial and I'm not sure if it can be done because the copyright but I want to open the discussion about it.

In this PR I am copying supervisor3 from klarna to brod and renaming it. The main reason is that supervisor3 has bug that does supervisor3 to stuck for ever and the [fix hasn't been merged](https://github.com/klarna/supervisor3/pull/24) after 2 years.

The main reason is that the author didn't sign the CLA. The [last comment](https://github.com/klarna/supervisor3/pull/24#issuecomment-815043986) from @zmstone is about polling the changes to kafka4beam/supervisor3, but I have the impression that it would mean to publish a new package and rename the module so it doesn't crash with supervisor3 if the clients are using that package too. Hence, I am coping the module inside brod, renaming it and apply the fix proposed in [the pull request](https://github.com/klarna/supervisor3/pull/24).

We have this change in our production environment for 2 years and it fixed the stuck problem so I think the rest of brod clients will like to see the fix in master too.

Let me know what you think.

(Full credits to [egertak](https://github.com/egertak) for fixing the bug)


zmstone (approved):

id (approved):

rebar.lock
```diff
@@ -1,14 +1,12 @@
...
 [{<<"crc32cer">>,{pkg,<<"crc32cer">>,<<"0.1.8">>},1},
  {<<"kafka_protocol">>,{pkg,<<"kafka_protocol">>,<<"4.1.2">>},0},
- {<<"snappyer">>,{pkg,<<"snappyer">>,<<"1.2.9">>},0},
- {<<"supervisor3">>,{pkg,<<"supervisor3">>,<<"1.1.12">>},0}]}.
+ {<<"snappyer">>,{pkg,<<"snappyer">>,<<"1.2.9">>},0}]}.
 [
 {pkg_hash,[
  {<<"crc32cer">>, <<"C6C2275C5FB60A95F4935D414F30B50EE9CFED494081C9B36EBB02EDFC2F48DB">>},
  {<<"kafka_protocol">>, <<"EE45DB88BD83526B1D3079DE19D5331AD7D755E549C00EC6F2CAA47C591E9588">>},
  {<<"snappyer">>, <<"9CC58470798648CE34C662CA0AA6DAAE31367667714C9A543384430A3586E5D3">>},
```
zmstone:
```suggestion
 {<<"snappyer">>, <<"9CC58470798648CE34C662CA0AA6DAAE31367667714C9A543384430A3586E5D3">>}]},
```

ieQu1 (approved):

gonzalobf:
Thanks, I don't have write access so please merge it when you think it is ready

zmstone:
Thank you. included in release 3.17.0

PR #562, opened 2023-07-17
fix: Extracts coveralls from `plugins` list to `project_plugins`

isaacsanders:
Coveralls isn't a plugin needed for the function of the library, so moving it to the `project_plugins` portion will prevent Elixir projects from making a network call when compiling to library after this is merged.

Currently this means we can't release new versions of our app that uses this library, but this change should be pretty straightforward to rubber stamp.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,7 +1,7 @@
 {deps, [ {kafka_protocol, "4.1.2"}
        , {snappyer, "1.2.9"}
        ]}.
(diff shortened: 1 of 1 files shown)


zmstone (approved):

zmstone:
Thank you @isaacsanders

zmstone:
released in 3.17.0

PR #589, opened 2024-06-29
Fix folding over transactions

### LINKED ISSUE #588: brod:fold doesn't work when there have been transactions
A failing test:

```erlang
t_fold_transactions(Config) when is_list(Config) ->
  Client = ?config(client),
  Topic = ?TOPIC,
  Partition = 0,
  Batch = [#{value => <<"one">>}, #{value => <<"two">>}],
  {ok, Tx} = brod:transaction(Client, <<"some_transaction">>, []),
  {ok, Offset} = brod:txn_produce(Tx, ?TOPIC, Partition, Batch),
  erlang:display("TXNPRODUCE"),
  erlang:display(Offset),
  ok = brod:commit(Tx),
  FoldF =
    fun F(#kafka_message{value = V}, Acc) -> {ok, F(V, Acc)};
        F(V, Acc) -> [V | Acc]
    end,
  FetchOpts = #{max_bytes => 100},
  ?assertMatch({_Result, O, reached_end_of_partition}
                when O =:= Offset + length(Batch) + 1,
    brod:fold(Client, Topic, Partition, Offset, FetchOpts, [], FoldF, #{})),
  ok.
```

Output:
```
%%% brod_consumer_SUITE ==> t_fold_transactions: FAILED
%%% brod_consumer_SUITE ==> {function_clause,
    [{lists,last,[[]],[{file,"lists.erl"},{line,228}]},
     {brod_utils,handle_fetch_rsp,7,
         [{file,"/home/indrek/gems/brod/src/brod_utils.erl"},{line,661}]},
     {brod_consumer_SUITE,t_fold_transactions,1,
         [{file,"/home/indrek/gems/brod/test/brod_consumer_SUITE.erl"},
          {line,443}]},
     {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1783}]},
     {test_server,run_test_case_eval1,6,
         [{file,"test_server.erl"},{line,1292}]},
     {test_server,run_test_case_eval,9,
         [{file,"test_server.erl"},{line,1224}]}]}
```

This is the line that throws an exception: https://github.com/kafka4beam/brod/blob/636e448f7d0183dc54fef31776cbe34d5d6f243b/src/brod_utils.erl#L644

The `brod_utils:fetch` ([flatten_batches](https://github.com/kafka4beam/brod/blob/636e448f7d0183dc54fef31776cbe34d5d6f243b/src/brod_utils.erl#L262)) seems to be control message aware, but at some point something gets lost.

I dug pretty deep, but I'm unsure how to fix this.

indrekj:
The brod:fold/8 did not work correctly when the last message in the message set a transaction commit.

The output looked like this:
```
%%% brod_consumer_SUITE ==> t_fold_transactions: FAILED
%%% brod_consumer_SUITE ==> {function_clause,
    [{lists,last,[[]],[{file,"lists.erl"},{line,228}]},
     {brod_utils,handle_fetch_rsp,7,
         [{file,"/home/indrek/gems/brod/src/brod_utils.erl"},{line,661}]},
     {brod_consumer_SUITE,t_fold_transactions,1,
         [{file,"/home/indrek/gems/brod/test/brod_consumer_SUITE.erl"},
          {line,443}]},
     {test_server,ts_tc,3,[{file,"test_server.erl"},{line,1783}]},
     {test_server,run_test_case_eval1,6,
         [{file,"test_server.erl"},{line,1292}]},
     {test_server,run_test_case_eval,9,
         [{file,"test_server.erl"},{line,1224}]}]}
```

The issue was that `#kafka_message{offset = LastOffset} = lists:last(Msgs),` was used, when the Msgs list was empty.

Now instead of trying to infer next offset from the kafka_message, we pass the NextFetchOffset from the brod_utils:fetch/4 function.

Fixes #588

### DIFF
--- a/src/brod_utils.erl
+++ b/src/brod_utils.erl
@@ -31,6 +31,7 @@
         , epoch_ms/0
         , fetch/4
         , fetch/5
+        , fetch_one_batch/4
         , fold/8
         , fetch_committed_offsets/3
         , fetch_committed_offsets/4
@@ -65,6 +66,8 @@
 
 -type req_fun() :: fun((offset(), kpro:count()) -> kpro:req()).
 -type fetch_fun() :: fun((offset()) -> {ok, {offset(), [brod:message()]}} |
+                                       {error, any()}).
+-type fetch_fun2() :: fun((offset()) -> {ok, {offset(), offset(), [brod:message()]}} |
                                        {error, any()}).
 -type connection() :: kpro:connection().
 -type conn_config() :: brod:conn_config().
@@ -331,7 +334,7 @@ fold(Client, Topic, Partition, Offset, Opts,
       ?BROD_FOLD_RET(Acc, Offset, {error, Reason})
   end;
 fold(Conn, Topic, Partition, Offset, Opts, Acc, Fun, Limits) ->
-  Fetch = make_fetch_fun(Conn, Topic, Partition, Opts),
+  Fetch = make_fetch_fun2(Conn, Topic, Partition, Opts),
   Infinity = 1 bsl 64,
   EndOffset = maps:get(reach_offset, Limits, Infinity),
   CountLimit = maps:get(message_count, Limits, Infinity),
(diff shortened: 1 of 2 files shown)


zmstone:
Hi @indrekj 
Thank you for the PR.

`make_fetch_fun` is exported, this change is technically breaking the compatibility for this function.
maybe it makes sense to export a `make_fetch_fun2` and a `fetch/5` 

`fetch/5` is a higher order of `fetch/4` which tries to fetch a non-empty batch.

indrekj:
@zmstone I pushed the changes. I'm unsure if I understood everything, but if you want any other changes, please let me know.

src/brod_utils.erl
```diff
@@ -443,7 +460,7 @@ do_fetch_committed_offsets(Conn, GroupId, Topics) when is_pid(Conn) ->
 %% @doc Fetch a message-set. If the given MaxBytes is not enough to fetch a
 %% single message, expand it to fetch exactly one message
 -spec fetch(connection(), req_fun(), offset(), kpro:count()) ->
-               {ok, {offset(), [brod:message()]}} | {error, any()}.
+               {ok, {offset(), offset(), [brod:message()]}} | {error, any()}.
 fetch(Conn, ReqFun, Offset, MaxBytes) ->
```
zmstone, line 458: I mean something like this:
```
fetch(Conn, ReqFun, Offset, MaxBytes, AllowEmptyBatch) ->
  case do_fetch(Conn, ReqFun, Offset, MaxBytes) of
    {ok, {StableOffset, NextOffset, []}} when not AllowEmptyBatch andalso NextOffset < StableOffset ->
      fetch(Conn, ReqFun, NextOffset, MaxBytes, AllowEmptyBatch);
    Other ->
      Other
end.

fetch(Conn, ReqFun, Offset, MaxBytes) ->
  case do_fetch(Conn, ReqFun, Offset, MaxBytes) of
    {ok, {StableOffset, _NextOffset, Msgs}} ->
      {ok, {StableOffset, Msgs}}; %% for backward compatibility
    Other ->
      Other
end.
```

so the caller will not have to deal with empty batches

indrekj: To clarify, there already is brod_utils:fetch/5 that is also exported:
```
fetch(Hosts, Topic, Partition, Offset, Opts) when is_list(Hosts) ->
fetch({Hosts, ConnCfg}, Topic, Partition, Offset, Opts) ->
fetch(Client, Topic, Partition, Offset, Opts) when is_atom(Client) ->
fetch(Conn, Topic, Partition, Offset, Opts) ->
```
Under the hood, it uses make_fetch_fun, which uses fetch/4.

Is it okay to add a new fetch/5 with quite a different signature? I want to be sure, as I don't write erlang daily. Also, do you think AllowEmptyBatch should be replaced with a list of options instead of a boolean flag? I'm okay with doing it either way.

zmstone: Right, sorry for missing that.
maybe `fetch_one_batch` ?

indrekj: @zmstone I got distracted, but finally addressed your comments

zmstone (approved):

zmstone:
thank you. published 4.1.0

PR #590, opened 2024-07-01
Fix type spec in brod_transaction

indrekj:
`brod:transaction` spec is:

```erlang
-spec transaction(client(), transactional_id(), transaction_config()) -> {ok, transaction()}.
```

and brod_transaction:init also has:

```erlang
init({Client, TxId, PropListConfig}) ->
  ClientPid = pid(Client),
```

but new and start_link were expecting only a pid. The client() spec itself is:

```erlang
-type client() :: client_id() | pid().
```

### DIFF
--- a/src/brod_transaction.erl
+++ b/src/brod_transaction.erl
@@ -133,10 +133,10 @@
 %%==============================================================================
 
 %% @see start_link/3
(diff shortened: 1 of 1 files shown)


zmstone:
Thank you for the PR @indrekj

zmstone:
thank you. this is included in 4.1.0

PR #591, opened 2024-07-02
Expose ack_fun through the init_info

indrekj:
`commit_fun` is already exposed, but `ack_fun` is not. This is useful when acking does not happen inside the worker process.

E.g. We divide messages between multiple processes and later in a separate process we create a transaction and commit a batch. Now we need to let the consumer know that it's okay to fetch more messages.

I briefly considered adding tests for this but wasn't sure how to do it. I've manually tested it, and it works. It follows the same pattern as commit_fun.


zmstone:
thank you for the pr.

zmstone:
thank you. included in release 4.1.0

PR #608, opened 2024-11-16
feat: add `get_coordinator` function in group subscriber v2

yastanotheruser:
Add an API function to get the group subscriber process coordinator PID. This allows to dynamically change the topics the group subscribes to by calling `brod_group_coordinator:update_topics`.

### DIFF
--- a/src/brod_group_subscriber_v2.erl
+++ b/src/brod_group_subscriber_v2.erl
@@ -35,6 +35,7 @@
         , start_link/1
         , stop/1
         , get_workers/1
+        , get_coordinator/1
         ]).
 
 %% brod_group_coordinator callbacks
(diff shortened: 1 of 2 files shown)


zmstone:
to lower the chance of race condition, maybe delete the gen_call to group subscriber v2.

yastanotheruser:
sorry I'm a bit slow. how could I read the state of the subscriber without a call to the gen_server?

zmstone:
> sorry I'm a bit slow. how could I read the state of the subscriber without a call to the gen_server?

hm.... my bad, I must have lost a few lines when copy-paste.
The race condition (although very unlikely to happen) is:

- Application code calls get_coordinator
- The coordinator process has to restart
- Application code calls update_topics

If what you want is to be able to call `brod_coordinator:update_topics`, I would suggest to expose `brod_group_subscriber_v2:update_topics` which delegates the `gen_cast` to coordinator process.

And add a test case to verify that new subscribers will be started and messages will be fetched after the function is called.

zmstone:
We can still add `get_coordinator` API, maybe it can be useful for some other purposes too.

PR #611, opened 2024-11-25
Exit group coordinator if member pid gone

### LINKED ISSUE #607: Process abnormal exit during termination when custom partitioning strategy is used
I have implemented a consumer using `brod_group_subscriber_v2` and with a custom partitioning strategy. The setup includes multiple consumers on different nodes. The setup works well until you try to shut down the application node by node.

The process crashes with the following error:
```erlang
(ErlangError) Erlang error: {:noproc, {:gen_server, :call, [#PID<0.4314.0>, {:assign_partitions, ...}]}}
  File "gen_server.erl", line 385, in :gen_server.call/3
  File "/app/deps/brod/src/brod_group_coordinator.erl", line 837, in :brod_group_coordinator.assign_partitions/1
  File "/app/deps/brod/src/brod_group_coordinator.erl", line 669, in :brod_group_coordinator.sync_group/1
  File "/app/deps/brod/src/brod_group_coordinator.erl", line 571, in :brod_group_coordinator.do_stabilize/3
  File "/app/deps/brod/src/brod_group_coordinator.erl", line 572, in :brod_group_coordinator.do_stabilize/3
  File "/app/deps/brod/src/brod_group_coordinator.erl", line 416, in :brod_group_coordinator.handle_info/2
  File "gen_server.erl", line 1123, in :gen_server.try_dispatch/4
  File "gen_server.erl", line 1200, in :gen_server.handle_msg/6
```

I wanted to ask whether this is an expected behavior? It seems as if the coordinator is still up but the `brod_group_subscriber_v2` process has already exited so it cannot respond. 

I am not quite sure how to fix it because the coordinator seems to be linked to the group subscriber. So I would assume this shouldn't happen at all?

If you have any suggestions, on how to avoid this crash, I could help with implementing it and opening a PR.

fmcgeough:
Fixes #607.

Terminate Group Coordinator if MemberPid is no longer alive when assignments_revoked, assign_partitions
or assignments_received must be called.

This is a first attempt to explore how to fix the issue raised in #607.

### DIFF
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -462,6 +462,13 @@ terminate(Reason, #state{ connection = Connection
 
 %%%_* Internal Functions =======================================================
 
+-spec ensure_member_pid_alive(pid()) -> ok 
+  ensure_member_pid_alive(MemberPid) ->
+    case brod_utils:is_pid_alive(MemberPid) of 
+      true -> ok
+      false -> exit(member_pid_shutdown)
+  end.
+
 -spec discover_coordinator(state()) -> {ok, state()}.
 discover_coordinator(#state{ client     = Client
                            , connection = Connection0
@@ -499,6 +506,7 @@ stabilize(#state{ rejoin_delay_seconds = RejoinDelaySeconds
     log(State0, info, "re-joining group, reason:~p", [Reason]),
 
   %% 1. unsubscribe all currently assigned partitions
+  ok = ensure_member_pid_alive(MemberPid), 
   ok = MemberModule:assignments_revoked(MemberPid),
 
   %% 2. some brod_group_member implementations may wait for messages
@@ -674,6 +682,7 @@ sync_group(#state{ groupId       = GroupId
   %% get my partition assignments
   Assignment = kpro:find(assignment, RspBody),
   TopicAssignments = get_topic_assignments(State, Assignment),
+  ok = ensure_member_pid_alive(MemberPid),
   ok = MemberModule:assignments_received(MemberPid, MemberId,
                                          GenerationId, TopicAssignments),
   NewState = State#state{is_in_group = true},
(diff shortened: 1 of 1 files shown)


zmstone:
Thank you for the PR @fmcgeough 
But this will not eliminate the race condition completely.
this should:
```
try
  MemberModule:assignments_revoked(MemberPid)
catch
  exit:{noproc,{gen_server,call,[MemberPid | _]}}} ->
    exit({shutdown, member_down})
end,
```

to avoid having to repeat for all three calls, a macro can be better:
```
-define(CALL_MEMBER(MemberPid, EXPR),
  try
    EXPR
  exit:{noproc,{gen_server,call,[MemberPid | _]}}} ->
    exit({shutdown, member_down})
  end).
```
and use it like this: `?CALL_MEMBER(MemberPid, MemberModule:assignments_revoked(MemberPid))`

src/brod_group_coordinator.erl
```diff
@@ -462,6 +462,15 @@ terminate(Reason, #state{ connection = Connection
 
 %%%_* Internal Functions =======================================================
 
+% Exit the process if MemberPid is set to pid but it's not
+% currently alive
+-spec ensure_member_pid_alive(pid()) -> ok.
+ensure_member_pid_alive(MemberPid) ->
+  case brod_utils:is_pid_alive(MemberPid) of
+    true -> ok;
+    false -> exit(member_pid_shutdown)
```
zmstone: should exit with `{shutdown, Reason}` otherwise there is still going to be a crash log.

fmcgeough: That makes sense to me. I'll modify the PR.

fmcgeough:
I've modified the PR to catch the exception rather than attempting to check if the process is alive. I'd appreciate any suggestions on how to do unit testing for this change.

zmstone (approved):

zmstone:
Thank you for the PR @fmcgeough

PR #634, opened 2025-06-18
Support update multiple topics by brod client

wiserfz:
Hi, first of all thank you for the work.

For now the `brod_client:get_metadata` and `brod_client:get_metadata_safe` function only support one topic, this PR modify those functions and support update multiple topics metadata.

Any feedback appreciated.

PR #623, opened 2025-03-20
Support Create Partitions

BlueCollarChris:
Adds the ability to create partitions for an existing topic (`brod:create_partitions/3,4`), exposing `create_partitions` from the underlying `kafka_protocol`.

Adds an optional client config `metadata_refresh_interval_seconds`: when set, the client starts a ticker in the `brod_client` process that periodically refreshes metadata for the topics it has producers running for, so producers are started automatically for newly discovered partitions (via the existing `maybe_start_partition_producer` path added in #636).

Reworked per review: the `brod_partitions_sync` module is gone in favor of the in-client ticker, the branch is rebased on current master with the formatting noise dropped, and `brod_SUITE` gains `t_auto_start_producers_for_new_partitions` which creates a topic, starts a producer, adds a partition, and asserts the new partition producer starts automatically.

PR #636, opened 2025-06-25
Start partition producer in get metadata function

### LINKED ISSUE #635: Produce message to increased partition failed
### Description

There is a problem when I get metadata by `brod_client:get_metadata` function and calculate a partition index which I need to send a message; and the partitions may be increased and the brod client doesn't support start a process for those new partitions.

### Solution

Create new supervisor process for those new partitions in `do_get_metadata` function.

wiserfz:
Close #635

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+- 4.4.5
+  - Start supervisor process for the new increase partitions at `do_get_metadata` function.
+
 - 4.4.4
   - Fixed `ListGroups` API request for Kafka Protocol API version 3.
 
--- a/src/brod_client.erl
+++ b/src/brod_client.erl
@@ -623,6 +623,7 @@ do_get_metadata(FetchMetadataFor, Topics,
                 #state{ client_id   = ClientId
                       , workers_tab = Ets
                       , config = Config
+                      , producers_sup = ProducersSup
                       } = State0) ->
   FetchTopics = case FetchMetadataFor of
              all -> all; %% in case no topic is given, get all
@@ -633,11 +634,18 @@ do_get_metadata(FetchMetadataFor, Topics,
   Request = brod_kafka_request:metadata(Conn, FetchTopics),
   case request_sync(State, Request) of
     {ok, #kpro_rsp{api = metadata, msg = Metadata}} ->
-      TopicMetadataArray = kf(topics, Metadata),
+      TopicsMetadata = kf(topics, Metadata),
+      ok = maybe_start_partition_producer(
+        filter_topics(TopicsMetadata),
+        brod_producers_sup:count_started_children(ProducersSup),
+        Topics,
+        ProducersSup
+      ),
+
       UnknownTopicCacheTtl = config(unknown_topic_cache_ttl, Config,
                                     ?DEFAULT_UNKNOWN_TOPIC_CACHE_TTL),
-      ok = update_partitions_count_cache(Ets, TopicMetadataArray, UnknownTopicCacheTtl),
-      ok = maybe_cache_unknown_topic_partition(Ets, Topics, TopicMetadataArray,
+      ok = update_partitions_count_cache(Ets, TopicsMetadata, UnknownTopicCacheTtl),
+      ok = maybe_cache_unknown_topic_partition(Ets, Topics, TopicsMetadata,
                                                UnknownTopicCacheTtl),
       {{ok, Metadata}, State};
     {error, Reason} ->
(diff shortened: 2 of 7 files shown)


wiserfz:
@zmstone hi, how about this PR

zmstone:
@wiserfz 

thank you  for the pr. i just got back from vacation last week, slowly catching up with the missed activities.
I'll have a closer look soon during this week.

src/brod_client.erl
```diff
@@ -981,6 +1000,74 @@ ensure_partition_workers(TopicName, State, F) ->
...
+    end,
+    lists:seq(OldPartitionCnt, OldPartitionCnt + (NewPartitionCnt - OldPartitionCnt) - 1)
+  ),
+
+  maybe_start_partition_producer(TopicsMetadata, Producers, Rest, ProducerSup, Configs);
+maybe_start_partition_producer(TopicsMetadata, Producers, [_Topic | Rest], ProducerSup, Configs) ->
+  maybe_start_partition_producer(TopicsMetadata, Producers, Rest, ProducerSup, Configs).
+
+-spec filter_topics([kpro:struct()]) -> #{topic() => non_neg_integer()}.
+filter_topics(TopicsMetadataArray) ->
```
zmstone, line 1041: I think we only get metadata for one topic at a time. i.e. this is a single element array.

zmstone: sorry, i was wrong. the call to get_metdata supports topics array.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+- 4.4.4
+  - Start supervisor process for the new increase partitions at `do_get_metadata` function.
```
zmstone, line 4: if the goal is to automatically start producers for newly discovered partitions, maybe brod_client should start a timer to periodically refresh partition counts for each topic which has a `producers_sup` started

src/brod_client.erl
wiserfz: yes, PR of #634 modified this function to supports topics array.

CHANGELOG.md
wiserfz: At the beginning of this, I do really consider of this, maybe should start a process and automatically get kafka metadata then start new partition producer, and also I noticed the PR of #623 which already implemented this feature.

But, there are two reasons lead me open this PR:

1. The bug of function `brod_client:get_metadata/2`, this function only modify the number of topic partition in ETS table, but not start the partition producer.
2. In my situation, I get kafka metadata and calculate the partition index which I need to produce the message; use #623 feature, there may be a jet lag that the partition producer is not start when I need to produce.

Of course, #623 feature is good and I think it's not conflict with this PR.

What do you think?

src/brod_client.erl
```diff
@@ -414,7 +415,15 @@ handle_call({get_transactional_coordinator, TransactionId}, _From, State) ->
   {Result, NewState} = do_get_transactional_coordinator(State, TransactionId),
   {reply, Result, NewState};
 handle_call({start_producer, TopicName, ProducerConfig}, _From, State) ->
-  {Reply, NewState} = do_start_producer(TopicName, ProducerConfig, State),
+  %% NOTE: Store ProducerConfig into the config in state
```
zmstone: producer config is per topic, it is already stored in supervisor child spec.
we can port get_childspec to brod_supervisor3

https://github.com/erlang/otp/blob/ec6323dd5065203737b0a48d074f116f0c1a2719/lib/stdlib/src/supervisor.erl#L752-L759

CHANGELOG.md
zmstone: yes, i think i can agree to this.
but should also stop producers if some partitions are deleted ?

src/brod_client.erl
```diff
@@ -981,6 +1000,74 @@ ensure_partition_workers(TopicName, State, F) ->
...
+    {TopicName, erlang:length(Partitions)}
+    || #{name := <<TopicName/binary>>, partitions := Partitions} <- NoErrors1,
+    is_list(Partitions)
+  ],
+
+  maps:from_list(NoErrors).
+
+-spec started_producers(pid()) -> #{topic() => non_neg_integer()}.
+started_producers(ProducerSup) ->
+  Producers = brod_supervisor3:which_children(ProducerSup),
```
zmstone: `brod_supervisor3:which_children(ProducerSup)` will give you the per-topic supervisors which supervises the per-partition workers

i think you'll need to do it like this:

```
started_producers(ProducerSup) ->
  TopicSups = which_producers(ProducerSup),
  lists:foldl(
    fun({Topic, Pid, _Type, _Mods}, Acc) when is_pid(Pid) ->
      PartitionWorkers = which_producers(Pid),
      case length(PartitionWorkers) of
        0 -> Acc;
        L -> Acc#{Topic => L}
    (_) ->
      Acc
    end, TopicSups).
```
and this
```
which_producers(Sup) ->
  try
    brod_supervisor3:which_children(Sup)
  catch
    _:_ ->
     []
  end.
```

zmstone (commented):
please kindly add some tests

CHANGELOG.md
wiserfz: > yes, i think i can agree to this. but should also stop producers if some partitions are deleted ?

kafka topic partitions only can be increased, it's can't be [reduced](https://kafka.apache.org/documentation/#basic_ops_modify_topic), so it doesn't need to be stopped.

src/brod_client.erl
wiserfz: Done.

src/brod_client.erl
wiserfz: Done.

wiserfz:
> please kindly add some tests

Done.

CHANGELOG.md
zmstone: yes, i'm referring to the case when a topic is deleted then created.

zmstone (approved):
this is good to merge, however, it would be nice to have another PR to address decreased partition count.

wiserfz:
Thank you for merge this PR.

Would you release a tag for this feature?

zmstone:
released 4.4.5

CHANGELOG.md
BlueCollarChris: Just getting back around to #623, will get it moving again sorry for the delay

PR #652, opened 2025-11-19
251119 consumer group health check

zmstone:
expose an API to check group consumer member's health status.

Note: one cannot check leader connectivity from `brod_client:get_leader_connection/3` because the topic-partition consumer by default starts its exclusive connection since version 4.2.0. `brod_client:get_leader_connection/3` will only establish a connection then continue to idle unless `share_leader_conn` is set to `true`.

### DIFF
--- a/src/brod_consumer.erl
+++ b/src/brod_consumer.erl
@@ -300,7 +300,7 @@ debug(Pid, File) when is_list(File) ->
 get_connection(Pid) ->
   gen_server:call(Pid, get_connection).
 
-%% @doc Return 'true' if the payload connection is alive.
+%% @doc Return 'ok' if the payload connection is alive.
 -spec check_connectivity(pid(), pos_integer()) -> ok | {error, term()}.
 check_connectivity(Pid, Timeout) ->
   try
--- a/src/brod_consumer.erl
+++ b/src/brod_consumer.erl
@@ -44,6 +44,7 @@
         , stop_maybe_kill/2
         , subscribe/3
         , unsubscribe/2
+        , check_connectivity/2
         ]).
 
 %% Debug API
@@ -299,6 +300,21 @@ debug(Pid, File) when is_list(File) ->
 get_connection(Pid) ->
   gen_server:call(Pid, get_connection).
 
+%% @doc Return 'true' if the payload connection is alive.
+-spec check_connectivity(pid(), pos_integer()) -> ok | {error, term()}.
+check_connectivity(Pid, Timeout) ->
+  try
+    case gen_server:call(Pid, is_connected, Timeout) of
+      true ->
+        ok;
+      false ->
+        {error, disconnected}
+    end
+  catch
+    exit : Reason ->
+      {error, Reason}
+  end.
+
 %%%_* gen_server callbacks =====================================================
 
 init({Bootstrap0, Topic, Partition, Config}) ->
@@ -402,6 +418,8 @@ handle_info(Info, State) ->
   {noreply, State}.
 
 %% @private
+handle_call(is_connected, _From, #state{connection = C} = State) ->
+  {reply, is_pid(C) andalso is_process_alive(C), State};
 handle_call(get_connection, _From, #state{connection = C} = State) ->
   {reply, C, State};
 handle_call({subscribe, Pid, Options}, _From,
--- a/src/brod_group_subscriber_v2.erl
+++ b/src/brod_group_subscriber_v2.erl
@@ -209,8 +209,8 @@ get_workers(Pid) ->
 get_workers(Pid, Timeout) ->
   gen_server:call(Pid, get_workers, Timeout).
 
-%% @doc Returns empty list if all workers are healthy,
-%% otherwise a list of errors.
+%% @doc Returns `ok' if all workers are healthy,
+%% otherwise `{error, Reasons}' where `Reasons' is a list of errors.
 -spec health_check(pid(), timeout()) -> ok | {error, [map()]}.
 health_check(Pid, Timeout) ->
   try
--- a/src/brod_group_subscriber_v2.erl
+++ b/src/brod_group_subscriber_v2.erl
@@ -35,6 +35,7 @@
         , start_link/1
         , stop/1
         , get_workers/1
+        , health_check/2
         ]).
 
 %% brod_group_coordinator callbacks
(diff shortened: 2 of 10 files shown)


src/brod_consumer.erl
```diff
@@ -299,6 +300,21 @@ debug(Pid, File) when is_list(File) ->
 get_connection(Pid) ->
   gen_server:call(Pid, get_connection).
 
+%% @doc Return 'true' if the payload connection is alive.
```
thalesmg:
```suggestion
%% @doc Return 'ok' if the connection is alive.
```

src/brod_consumer.erl
```diff
@@ -299,6 +300,21 @@ debug(Pid, File) when is_list(File) ->
...
+check_connectivity(Pid, Timeout) ->
+  try
+    case gen_server:call(Pid, is_connected, Timeout) of
+      true ->
+        ok;
+      false ->
+        {error, disconnected}
+    end
+  catch
+    exit : Reason ->
```
thalesmg, line 314: nit: `noproc` could be seen as `disconnected` too

src/brod_group_subscriber_v2.erl
```diff
@@ -208,6 +209,22 @@ get_workers(Pid) ->
 get_workers(Pid, Timeout) ->
   gen_server:call(Pid, get_workers, Timeout).
 
+%% @doc Returns empty list if all workers are healthy,
```
thalesmg:
```suggestion
%% @doc Returns `'ok'` if all workers are healthy,
```

src/brod_topic_subscriber.erl
```diff
@@ -270,6 +271,17 @@ stop(Pid) ->
...
 
+%% @doc Returns empty if all consumers are connected,
+%% otherwise a list of errors.
+-spec health_check(pid(), pos_integer()) -> [map()].
+health_check(Pid, Timeout) ->
+  try
+    gen_server:call(Pid, {health_check, Timeout}, Timeout + 50)
+  catch
+    exit : Reason ->
+      ?MF_ERR_MAP(Reason)
```
thalesmg:
```suggestion
      [?MF_ERR_MAP(Reason)]
```

test/brod_group_subscriber_SUITE.erl
```diff
@@ -308,6 +310,44 @@ t_get_workers(Config) when is_list(Config) ->
...
+      Timeout = 5000,
+      ?check_trace(
+         #{ timeout => 10000 },
+         %% Run stage:
+         begin
+           {ok, SubscriberPid} = start_subscriber(?group_id, Config, [Topic], InitArgs),
+           %% Send a message to ensure workers are established
+           produce({Topic, Partition}, <<"test">>),
+           {ok, _} = ?wait_message(Topic, Partition, <<"test">>, _),
+           %% Test 1: When all workers are healthy, should return empty list
```
thalesmg:
```suggestion
           %% Test 1: When all workers are healthy, should return `ok`
```

test/brod_group_subscriber_SUITE.erl
```diff
@@ -308,6 +310,44 @@ t_get_workers(Config) when is_list(Config) ->
...
+           %% Test 2: Test with non-existent PID
+           FakePid = list_to_pid("<0.999.0>"),
+           HealthResult2 = brod_group_subscriber_v2:health_check(FakePid, 1000),
+           %% Cleanup: stop the subscriber
+           ok = stop_subscriber(Config, SubscriberPid),
+           {HealthResult1, HealthResult2}
+         end,
+         %% Check stage:
+         fun({HealthResult1, HealthResult2}, _Trace) ->
+             %% When healthy, should return empty list
```
thalesmg:
```suggestion
             %% When healthy, should return `ok`
```

src/brod_consumer.erl
zmstone: noproc is different, likely a race condition during shutdown, reporting disconnected could be confusing.

src/brod_group_subscriber_v2.erl
```diff
@@ -545,6 +564,25 @@ stop_consumers(Config) ->
     end,
     Topics).
 
+do_health_check(Workers0, Timeout) ->
```
qzhuyan, line 567: Should handle Workers0 = []?

zmstone: you mean `#{}` (empty map) ?

it can happen when there is no partition assigned to this consumer when there are more consumers than partitions.

src/brod_utils.erl
```diff
@@ -985,6 +986,50 @@ is_batch(_) -> false.
...
+  Collector = erlang:spawn_link(fun() -> collect_pmap_results(Parent, SeqL, []) end),
+  Workers = lists:map(fun({Seq, Item}) ->
+                        erlang:spawn_link(fun() -> Collector ! {Seq, Fun(Item)} end)
+                      end, lists:zip(SeqL, List)),
+  receive
+    {Collector, Res} ->
+      ok = drain_exits([Collector | Workers]),
+      Res
+  after Timeout ->
+    lists:foreach(fun unlink_kill/1, [Collector | Workers]),
```
qzhuyan, line 1005: So the following drain_exits are for the best efforts? as unlink then kill won't promise a EXIT message.

src/brod_group_subscriber_v2.erl
qzhuyan: yes empty map. Ok so that is ok to mark health check success.

qzhuyan (approved):

src/brod_utils.erl
```diff
@@ -985,6 +986,50 @@ is_batch(_) -> false.
 nolink(C) when is_list(C) -> [{nolink, true} | C];
 nolink(C) when is_map(C) -> C#{nolink => true}.
 
+%% @doc Parallel map with timeout.
+-define(IS_TIMEOUT(T), ((is_integer(Timeout) andalso Timeout >= 0) orelse Timeout =:= infinity)).
+-spec pmap(fun((A) -> B), list(A), timeout()) -> list(B).
+pmap(_Fun, [], _Timeout) -> [];
+pmap(Fun, [_|_] = List, Timeout) when is_function(Fun, 1), ?IS_TIMEOUT(Timeout) ->
+  SeqL = lists:seq(1, length(List)),
```
qzhuyan, line 994: would prefer to have some reference() + seq to avoid message collisions?

zmstone: short-lived process is spawned as worker and collector, so we don't expect any other kind of messages.

src/brod_utils.erl
```diff
@@ -985,6 +986,50 @@ is_batch(_) -> false.
...
+  Workers = lists:map(fun({Seq, Item}) ->
+                        erlang:spawn_link(fun() -> Collector ! {Seq, Fun(Item)} end)
+                      end, lists:zip(SeqL, List)),
+  receive
+    {Collector, Res} ->
+      ok = drain_exits([Collector | Workers]),
+      Res
+  after Timeout ->
+    lists:foreach(fun unlink_kill/1, [Collector | Workers]),
+    ok = drain_exits([Collector | Workers]),
```
qzhuyan, line 1006: there is a tiny chance, `{Collector, Res}` in delivered in mailbox when handling timeout, I think. so flush that again?

qzhuyan: maybe easier just let Collector handles the timeout.

PR #660, opened 2026-05-15
bug: OffsetFetch -1 silently drops partition, begin_offset falls back to undefined

flysen:
When Kafka's OffsetFetch returns committed_offset=-1 with error_code=NONE, it signals that no committed offset exists — e.g. the group is new, the topic was deleted and recreated, or offsets.retention.minutes expired.

**The bug**

In get_committed_offsets/2, when Offset0 =:= -1, the partition is silently dropped from the result. resolve_begin_offsets/3 then returns begin_offset=undefined, and brod_group_subscriber falls back to consumer config (typically latest) with no error, no log, no alert. offset_reset_policy is never consulted.

**Fix direction**

In get_committed_offsets/2, when Offset0 =:= -1, add {{Topic, Partition}, {error, no_committed_offset}} to the accumulator and handle it in brod_group_subscriber like OFFSET_OUT_OF_RANGE.

**This PR**

Adds a test that demonstrates the gap. The test is currently skipped (known bug). Removing the skip is the acceptance criteria for the fix.

### DIFF
--- a/test/brod_group_coordinator_SUITE.erl
+++ b/test/brod_group_coordinator_SUITE.erl
@@ -39,6 +39,7 @@
 %% Test cases
 -export([ t_acks_during_revoke/1
         , t_update_topics_triggers_rebalance/1
+        , t_offset_fetch_minus_one_silently_falls_back_to_begin_offset/1
         ]).
 
 -define(assert_receive(Pattern, Return),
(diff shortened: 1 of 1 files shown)


zmstone:
Hi @flysen 
Thank you for the PR.
I pulled your commit into #661

PR #665, opened 2026-06-11
fix(client): break brod_client <-> producers_sup deadlock

### LINKED ISSUE #662: Deadlock between brod_client and brod_producers_sup, especially since 4.4.5
Hi!

https://github.com/kafka4beam/brod/pull/636 added useful functionality to auto-start producers when producer count increased.
Unfortunately, the implementation increases risk of an infinite deadlock, which I've observed at application start-up.

1. `brod_client` initializes (PID = 1)
2. `brod_client` starts a [producer supervisor (PID = 2)](https://github.com/kafka4beam/brod/blob/4.5.4/src/brod_client.erl#L364)
3. Concurrently
  a) `brod_producers_sup` (PID = 2) calls back into the parent `:brod_client` (PID = 1) to [get partition count](https://github.com/kafka4beam/brod/blame/4.5.4/src/brod_producers_sup.erl#L162)
  b) There is a call from arbitrary library user process (PID = 3) to `:brod_client.get_producer`, e.g. delegating from a `brod:produce` / `:brod.get_producer`
      1. `:brod_client.get_producer` does not find an existing producer and delegates to
      2. `:brod_client.maybe_start_producer` which calls [into PID = 1 with `{auto_start_producer, :infinity}`
](https://github.com/kafka4beam/brod/blob/4.5.4/src/brod_client.erl#L946).
      3. ☝️ call executes before 3. a) in `:brod_client` (PID = 1) and as partitions are [not cached](https://github.com/kafka4beam/brod/blob/4.5.4/src/brod_client.erl#L576), delegates to `do_get_metadata`.
      4. `do_get_metadata` [counts children of `ProducersSup` (PID = 2)](https://github.com/kafka4beam/brod/blob/4.5.4/src/brod_client.erl#L640), calling into PID = 2 with `gen_server:call(ProducersSup, :which_children, infinity).`

* After 3. a), `brod_producers_sup` (PID = 2) is waiting on `brod_client` (PID = 1) to get partitions count with timeout `:infinity`
* After 3. b) iv. `brod_client`(PID = 1) is waiting on `brod_producers_sup` (PID = 2) to count children with timeout `:infinity` due to a `get_metadata` call
* The `brod_client` process is infinitely deadlocked until operator intervention. No events can be consumer nor produced.

Per my understanding, risk of this deadlock between the `brod_client` and `brod_producers_sup` existed already before 4.4.5, but the risk is increased since 4.4.5 due to increased calls into `ProducersSup`.

Do you think that the count of partitions could be passed from `brod_client` to `brod_producers_sup`, avoiding a blocking call back into parent process? Alternatively, the call into the parent could be made non-blocking.

zmstone:
fixes https://github.com/kafka4beam/brod/issues/662

### DIFF
--- a/src/brod_producers_sup.erl
+++ b/src/brod_producers_sup.erl
@@ -159,7 +159,7 @@ init({?PARTITIONS_SUP, _ClientPid, _Topic, _Config}) ->
   post_init.
 
 post_init({?PARTITIONS_SUP, ClientPid, Topic, Config}) ->
-  case brod_client:get_partitions_count(ClientPid, Topic) of
+  case get_partitions_count(ClientPid, Topic) of
     {ok, PartitionsCnt} ->
       Children = [ producer_spec(ClientPid, Topic, Partition, Config)
                  || Partition <- lists:seq(0, PartitionsCnt - 1) ],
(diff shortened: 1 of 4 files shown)


src/brod_producers_sup.erl
```diff
@@ -173,6 +173,22 @@ post_init({?PARTITIONS_SUP, ClientPid, Topic, Config}) ->
...
+%% Resolve partition count without calling back into brod_client (issue #662).
+%% brod_client populates its partition count cache via
+%% `validate_topic_existence/3' immediately before starting this per-topic
+%% supervisor, so the cache lookup is expected to hit on the brod_client-driven
+%% code path. This avoids a synchronous `gen_server:call' into brod_client
+%% during `post_init', which would deadlock when brod_client is concurrently
+%% walking the producer supervisor tree from `count_started_children/1'
+%% inside `do_get_metadata'.
+%% Fall back to the synchronous brod_client API on cache miss to preserve
+%% behaviour for any external caller of `start_producer/4'.
```
thalesmg: not sure I understand this remark and fallback.

could you please expand on it?  i.e., how would this affect an external caller if it starts a child the normal way, in which case it is supposed to hit the cache according to the comment?

thalesmg (approved):

src/brod_producers_sup.erl
zmstone: the new commit refined this comment.

PR #668, opened 2026-07-20
OTP-29 fixes

baronbosse:
This removes all old style catch statements and uses the new syntax. Also a no longer allowed "setting of variable" inside a function call is removed.

Should now compile and test cleanly on OTP-29.


zmstone (approved):

zmstone:
thank you for the pr @baronbosse 
- tag: 4.5.7
- published to [hex.pm](https://hex.pm/packages/brod/4.5.7)

PR #669, opened 2026-07-27
Add full support for static membership (with heartbeat)

whatyouhide:
This is a neat change IMO because if we use the group instance ID (hence static membership), then with this PR we can avoid `LeaveGroup` when a consumer shuts down, resulting in proper no-rebalance semantics.

Thoughts?


zmstone:
thank you for the pr.

i'd like to know why the hardcoded static versions for
sync_group, offset_commit, and heartbeat.

IMO we should expand the supported versions range 

- sync_group: expand from `{0,0}` to at least `{0,3}`: https://github.com/kafka4beam/brod/blob/ef086a2de8217eacf7c12321a0c9d53c105cedab/src/brod_kafka_apis.erl#L151
- offset_commit: expand from `{2,2}` to at least `{2,3}`: https://github.com/kafka4beam/brod/blob/ef086a2de8217eacf7c12321a0c9d53c105cedab/src/brod_kafka_apis.erl#L145
- heartbeat: it is already `{0,4}` which covers the hardcoded version 3, i wonder why, does v4 have some issue? https://github.com/kafka4beam/brod/blob/ef086a2de8217eacf7c12321a0c9d53c105cedab/src/brod_kafka_apis.erl#L149

If the hardcoded versions are the minimal versions required to support static membership, the code shall check if the queried ranges include the minimal versions right after connected to decide if `static_membership = true` should be added to state record.

whatyouhide:
@zmstone I did another pass where we're trying to be more lenient and dynamic on the Kafka API versioning stuff you mentioned. I’m no Kafka expert so let me know if this is completely off base. Thanks for reviewing!

src/brod_group_coordinator.erl
```diff
@@ -75,6 +75,12 @@
         {lo_cmd_stabilize, AttemptCount, Reason}).
 
 -define(INITIAL_MEMBER_ID, <<>>).
+-define(STATIC_MEMBERSHIP_API_VERSIONS,
```
zmstone: these are lower boundaries of api version to for static membership please name this macro`STATIC_MEMBERSHIP_MIN_API_VERSIONS`.
and add a comment above to describe this macro and which Kafka version introduced static membership etc.

src/brod_kafka_apis.erl
```diff
@@ -70,6 +71,22 @@ default_version(API) ->
 pick_version(Conn, API) ->
   do_pick_version(Conn, API, supported_versions(API)).
 
+%% @doc Return true when both brod and the broker support an API version.
+-spec supports_version(conn(), api(), vsn()) -> boolean().
+supports_version(Conn, API, Vsn) ->
```
zmstone, line 76: i think it can be implemented as `do_pick_version(Conn, Api) >= Vsn`.

src/brod_group_coordinator.erl
```diff
@@ -446,6 +470,8 @@ handle_info({msg, _Pid, #kpro_rsp{ api = heartbeat
   EC = kpro:find(error_code, Body),
   State = State0#state{hb_ref = ?undef},
   case ?IS_ERROR(EC) of
+    true when EC =:= ?fenced_instance_id ->
+      {stop, ?fenced_instance_id, State};
```
zmstone: `fenced_instance_id` is a non-normal exit reason, so the linked member process dies too and its supervisor will typically restart it, and the restarted member re-joins with the same `group_instance_id`, fencing the other member right back. with two members misconfigured to share one instance id this becomes a supervisor-driven crash-restart ping-pong that is hard to diagnose (no rebalances, just silent slot takeovers).

instead of stopping, suggest to handle it like other recoverable errors: emit an error-level log hinting the likely cause (the same `group_instance_id` configured for another member of the groupk, e.g. if two Erlang nodes started with the same config), then start a timer to retry stabilization (e.g. the `?LO_CMD_STABILIZE` delay path) rather than crash. that keeps the conflict visible in logs and rate-limits the mutual fencing instead of turning it into a supervisor restart loop. in the legit takeover case (old member fenced while its replacement joins during e.g. a rolling restart) the old member is shutting down anyway, so a delayed retry is harmless there.

src/brod_group_coordinator.erl
```diff
@@ -385,19 +398,30 @@ init({Client, GroupId, Topics, Config, CbModule, MemberPid}) ->
...
-    catch throw : Reason ->
-      stabilize(State, 0, Reason)
+    catch
+      throw : ?fenced_instance_id ->
+        {stop, ?fenced_instance_id};
+      throw : Reason ->
+        stabilize(State, 0, Reason)
     end,
-  {noreply, NewState};
+  case CommitResult of
```
zmstone: the intermediate `CommitResult` + `case` can be avoided by returning from the try/catch directly:

```erlang
try
  {ok, NewState} = do_commit_offsets(State),
  {noreply, NewState}
catch
  throw : ?fenced_instance_id ->
    {stop, ?fenced_instance_id, State};
  throw : Reason ->
    {ok, NewState1} = stabilize(State, 0, Reason),
    {noreply, NewState1}
end;
```

src/brod_group_coordinator.erl
zmstone: this applies to all other stop and exit cases below.

src/brod_kafka_apis.erl
```diff
@@ -122,14 +140,16 @@ do_pick_version(Conn, API, {Min, Max} = MyRange) ->
 lookup_vsn_range(Conn, API) ->
   case ets:lookup(?ETS, Conn) of
     [] ->
-      case kpro:get_api_versions(Conn) of
```
zmstone, line 125: we should keep this.
if the reason is to support v7 offset commit by-passing the verified versions in kafka_protocol, then the order is wrong.

we'll have to fix kafka_protcol first to support v7 properly with test coverages etc.
then bump it to v7 in this repo.

I wonder if commit offset api v7 is a must for static member id. if not, that can be a different task.

zmstone (commented):
turns out my comments stayed in pending state for days.
sorry for the delay.

whatyouhide:
@zmstone I think I addressed all the feedback, and I added an entry in the CHANGELOG.

zmstone:
@whatyouhide  forgot to push ?

zmstone:
published 4.5.8, please rebase and add changelog to 4.6.0

whatyouhide:
@zmstone done ✔️

src/brod_group_coordinator.erl
```diff
@@ -579,19 +636,37 @@ stabilize(#state{ rejoin_delay_seconds = RejoinDelaySeconds
...
+retry_after_fencing(
+  #state{rejoin_delay_seconds = RejoinDelaySeconds} = State,
+  AttemptNo
+) ->
+  log(State, error,
+      "Static member was fenced by another member using the same "
+      "group_instance_id; check for duplicate group member configuration. "
+      "Retrying stabilization after ~p seconds.",
+      [RejoinDelaySeconds]),
+  _ = erlang:send_after(timer:seconds(RejoinDelaySeconds), self(),
```
zmstone: RejoinDelaySeconds is perhaps too short for fenced retry, the purpose is to raise awareness, we should perhaps use `max(RejoinDelaySeconds, 10)` for this timer.

Another thing is, the subscribers will continue to fetch and commit while waiting.
We should:
1. Revoke assignments immediately
2. Cancel offset commit timer

i.e. mirroring `stabilize` steps 1–2
-- this will require the handle of LO_CMD_STABILIZE adapted to skip step 1-2.

src/brod_group_coordinator.erl
```diff
@@ -75,10 +75,11 @@
         {lo_cmd_stabilize, AttemptCount, Reason}).
 
 -define(INITIAL_MEMBER_ID, <<>>).
--define(STATIC_MEMBERSHIP_API_VERSIONS,
+%% KIP-345 introduced static membership in Kafka 2.3. These are the minimum
+%% API versions whose requests carry the static member's group instance ID.
+-define(STATIC_MEMBERSHIP_MIN_API_VERSIONS,
         [ {join_group, 5}
         , {sync_group, 3}
-        , {offset_commit, 7}
```
zmstone: offset_commit is maybe not a hard requirement, but i just noticed that group_instance_id was never fed to `brod_kafka_request:offset_commit/2`

this means `throw : ?fenced_instance_id ->` is dead code for offset commit call/info handler

src/brod_kafka_apis.erl
```diff
@@ -70,6 +71,11 @@ default_version(API) ->
 pick_version(Conn, API) ->
   do_pick_version(Conn, API, supported_versions(API)).
 
+%% @doc Return true when both brod and the broker support an API version.
+-spec supports_version(conn(), api(), vsn()) -> boolean().
+supports_version(Conn, API, Vsn) ->
+  do_pick_version(Conn, API, supported_versions(API)) >= Vsn.
```
zmstone: need to catch `error : {unsupported_vsn_range, _, _, _}` to return `false`.

src/brod_group_coordinator.erl
zmstone: Add:
```
is_commit_possible(?illegal_generation) -> false;
is_commit_possible(?unknown_member_id) -> false;
is_commit_possible(?fenced_instance_id) -> false;
is_commit_possible(_) -> true.
```

CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
 # Changelog
 
+- 4.6.0
+  - Add KIP-345 **static group membership** for coordinators configured with an
```
zmstone, line 4: please also document the api range changes.
sync_group => {0, 3}
and maybe_send_heartbeat changed from hard coded v0 to negotiated version.

this typically works fine with Kafka, but there are Kafka compatible servers which might be sensitive to different version combinations. worth calling it out so users aware when upgrade.

src/brod_group_coordinator.erl
```diff
@@ -154,6 +161,10 @@
         , protocol_name                  :: protocol_name()
           %% Static member ID
         , group_instance_id              :: binary()
+          %% True when the caller supplies a stable group instance ID.
+        , static_membership_requested     :: boolean()
+          %% True when the broker supports every KIP-345 request version.
+        , static_membership               :: boolean()
```
zmstone: nit:
```suggestion
        , static_membership_requested    :: boolean()
          %% True when the broker supports every KIP-345 request version.
        , static_membership              :: boolean()
```

whatyouhide:
@zmstone did another pass at addressing the latest batch of comments.

zmstone:
thank you for the quick update.
would be nice if you can send another pr to expand offset commit request to the max supported by kafka_protocol.

zmstone (approved):

whatyouhide:
@zmstone any chance we could get a release out to Hex for this? 🙏

zmstone:
done. published 4.6.0

PR #670, opened 2026-08-06
Handle offset commits during a connection-down window

whatyouhide:
We've seen this happen in production a handful of times.

CleanShot 2026-08-06 at 10 13 36@2x

The fix here _should_ be, I think, to throw an error that gets caught and returned as a normal error. Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-- Unreleased
+- 4.5.8
   - Fix a consumer-group coordinator crash when offset commits race with
     recovery from a lost broker connection.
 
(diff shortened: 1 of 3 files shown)


zmstone:
thank you for the pr.

zmstone:
please help to update changelog.

whatyouhide:
@zmstone updated. I went with "Unreleased" as the version so you can update that to whatever version you'll release next once that release is ready to go; how does that sound?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+- Unreleased
```
zmstone: add it to 4.5.8, i'll release it after merged.
```suggestion
- 4.5.8
```

whatyouhide: @zmstone got it, done!

zmstone (approved):

Issue #503, opened 2022-05-11
Dispatch the metrics via telemetry


slashmili:
Hello!

I'm wondering if you are open to a contribution to introduce [telemetry](https://github.com/beam-telemetry/telemetry) into this project?

There lots of important metric that I'd like to tap into especially when consumer connect and rebalance and would be a great addition to this library.

zmstone:
Hi @slashmili

It'd have to be configurable though.
WDYT? @ieQu1 @qzhuyan @id 
and @kjellwinblad @starbelly @vikas15bhardwaj inputs are appreciated.

slashmili:
@zmstone configureable to disable it completely? Ok fair enough.

Do you prefer it to be done via Macros or via a condition that only runs during runtime? I'm asking since if Macro is they way to go, I need to do some study first.

zmstone:
Hi @slashmili 
Ideally the dependency should be optional for users who do not need beam-telemetry.
We can probably introduce a `brod_telemetry` module to confine the telemetry dependencies
and use a macro to control if it's a no-op or call telemetry instead.
e.g.
```
-ifndef(BROD_USE_TELEMETRY).
execute(_Name, _Measurements, _Metadata) -> ok.
-else.
execute(Name, Measurements, Metadata) -> telemetry:exeucte(Name, Measurements, Metadata).
-endif.
```

starbelly:
Telemetry in brod would be a big win 👍  💯

spencerdcarlson:
I'd love to help on this wherever I can. I think librdkafka's [STATISTICS](https://github.com/edenhill/librdkafka/blob/2aadc45974d32b7c7ff91805588be3fc9951f4ad/STATISTICS.md) page could be helpful when trying to figure out exactly what metrics should be emitted.

spencerdcarlson:
@zmstone regarding 
> Hi @slashmili Ideally the dependency should be optional for users who do not need beam-telemetry. 

Does this actually need to be the case? I thought by nature telemetry uses dynamic dispatching to delegate the execution of the handler or "work" process to the the subscriber so the framework has an "optional" built in. If you subscribe to the events then you "opt in" and some overhead is added, if you do not subscribe, then there is no handler for the given event and there is no overhead added. 

It looks like the only overhead added to brod would be an ets lookup to see if a handler for an event exists -- This seems like a very small cost to pay in favor of the added complexity of an optional dependency and macro.  
[...]

PR #671, opened 2026-08-14
Depend on telemetry and emit produce request event

zachallaun:
Add a dependency on the telemetry library and a brod_metrics module confining all telemetry calls, based on the pattern established in [kafka4beam/wolff](https://github.com/kafka4beam/wolff/blob/4feffa6534a960377507fb2dc840db07548b2533/src/wolff_metrics.erl).

The first event, `[brod, product_request_sent]`, is emitted by brod_producer for every produce request successfully sent on the wire. This makes the actual wire-level batching observable, which is otherwise invisible to callers

More events can be added to brod_metrics incrementally.

Related: #503, #512

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
-{deps, [{kafka_protocol, "4.3.5"}, {telemetry, "1.4.2"}]}.
+{deps, [{kafka_protocol, "4.3.5"}, {telemetry, "~> 1.0"}]}.
 {project_plugins, [{rebar3_lint, "~> 3.2.5"}]}.
 {edoc_opts, [{preprocess, true}]}.
 {erl_opts, [warnings_as_errors, warn_unused_vars,warn_shadow_vars,warn_obsolete_guard,debug_info]}.
--- a/rebar.config
+++ b/rebar.config
@@ -1,4 +1,4 @@
-{deps, [{kafka_protocol, "4.3.5"}]}.
+{deps, [{kafka_protocol, "4.3.5"}, {telemetry, "1.4.2"}]}.
 {project_plugins, [{rebar3_lint, "~> 3.2.5"}]}.
 {edoc_opts, [{preprocess, true}]}.
 {erl_opts, [warnings_as_errors, warn_unused_vars,warn_shadow_vars,warn_obsolete_guard,debug_info]}.
--- a/src/brod_metrics.erl
+++ b/src/brod_metrics.erl
@@ -35,19 +35,25 @@
 %% Emitted by producers for each produce request successfully sent on
 %% wire, i.e. one event per Kafka produce request (message batch).
 %%
+%% Note that retried batches emit the event again: the event counts
+%% wire sends, so a summed `count' can exceed the number of unique
+%% messages produced when retries occur.
+%%
 %% Measurements:
 %% <ul>
 %% <li>`count': number of messages in the request.</li>
 %% <li>`bytes': total size of keys, values and headers of the messages
-%%     in the request (before encoding and compression).</li>
+%%     in the request (before encoding and compression), plus 8 bytes
+%%     per message accounting for the timestamp.</li>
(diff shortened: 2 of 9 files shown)


rebar.config
```diff
@@ -1,4 +1,4 @@
-{deps, [{kafka_protocol, "4.3.4"}]}.
+{deps, [{kafka_protocol, "4.3.4"}, {telemetry, "1.4.2"}]}.
```
zmstone: Consider a range constraint such as `{telemetry, "~> 1.0"}` instead of an exact pin. telemetry is a dependency shared by many libraries, and an exact pin in a library causes version conflicts for consumers. Concrete case: wolff pins `{telemetry, "1.1.0"}`, so a project that depends on both brod and wolff would hit a hard conflict in Mix and a nearest-wins warning in rebar3.

src/brod_metrics.erl
```diff
@@ -0,0 +1,60 @@
...
+%% @doc Emit a `[brod, produce_request_sent]' event.
+%%
+%% Emitted by producers for each produce request successfully sent on
+%% wire, i.e. one event per Kafka produce request (message batch).
+%%
+%% Measurements:
+%% <ul>
+%% <li>`count': number of messages in the request.</li>
+%% <li>`bytes': total size of keys, values and headers of the messages
+%%     in the request (before encoding and compression).</li>
```
zmstone: `brod_utils:bytes/1` also adds 8-byte timestamp.

src/brod_metrics.erl
```diff
@@ -0,0 +1,60 @@
...
+-export([produce_request_sent/3]).
+
+-type batch_input() :: brod:batch_input().
+-type partition() :: brod:partition().
+-type topic() :: brod:topic().
+
+%% @doc Emit a `[brod, produce_request_sent]' event.
+%%
+%% Emitted by producers for each produce request successfully sent on
+%% wire, i.e. one event per Kafka produce request (message batch).
```
zmstone, line 36: worth noting here that retried batches emit the event again. The event counts wire sends, so a summed `count` can exceed the number of unique messages produced when retries occur.

src/brod_metrics.erl
```diff
@@ -0,0 +1,60 @@
...
+%% <li>`bytes': total size of keys, values and headers of the messages
+%%     in the request (before encoding and compression).</li>
+%% </ul>
+%%
+%% Metadata: `topic' and `partition' the request was sent to.
+-spec produce_request_sent(topic(), partition(), batch_input()) -> ok.
+produce_request_sent(Topic, Partition, BatchInput) ->
+  telemetry:execute([brod, produce_request_sent],
+                    #{ count => length(BatchInput)
+                     , bytes => brod_utils:bytes(BatchInput)
```
zmstone: nit: maybe worth exporting a function which folds the batch to return both count and bytes.

zachallaun: like a `batch_metadata` or similar?

zachallaun: (I mean measurements, not metadata)

rebar.config
zachallaun: Done!

src/brod_metrics.erl
zachallaun: Noted!

src/brod_metrics.erl
zachallaun: Noted!

src/brod_metrics.erl
zmstone: like `brod_utils:stats/1` which returns `{Count, Bytes}`.
and `brod_utils:bytes/1` calls `brod_utils:stats/1`, discard `Count`, returns `Bytes`.

zachallaun: Done!

zmstone (approved):

PR #672, opened 2026-08-17
Enable newer OffsetCommit request versions

whatyouhide:
Follow up to https://github.com/kafka4beam/kafka_protocol/pull/150 where we take advantage of that (and the new version of `kafka_protocol`) to enable newer request versions for `OffsetCommit`.

This in turn enables full support for static membership.

I added a new test file to have some more coverage, lmk if that's fine @zmstone.

### DIFF
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -918,7 +918,7 @@ do_commit_offsets_(#state{ groupId                  = GroupId
         PartitionOffset =
           [ {partition_index, Partition}
           , {committed_offset, Offset + 1} %% +1 since roundrobin_v2 protocol
-          , {committed_leader_epoch, -1}
+          , {committed_leader_epoch, -1} %% -1 = unknown leader epoch
           , {committed_metadata, Metadata}
           ],
         {Topic, PartitionOffset}
--- a/src/brod_group_coordinator.erl
+++ b/src/brod_group_coordinator.erl
@@ -83,6 +83,7 @@
 -define(STATIC_MEMBERSHIP_MIN_API_VERSIONS,
         [ {join_group, 5}
         , {sync_group, 3}
+        , {offset_commit, 7}
         , {heartbeat, 3}
         ]).
 -define(CALL_MEMBER(MemberPid, EXPR),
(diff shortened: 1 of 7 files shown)


src/brod_group_coordinator.erl
```diff
@@ -936,6 +939,7 @@ do_commit_offsets_(#state{ groupId                  = GroupId
     [ {group_id, GroupId}
     , {generation_id, GenerationId}
     , {member_id, MemberId}
+    , {group_instance_id, StaticMemberID}
     , {retention_time_ms, Retention}
```
zmstone, line 943: offset_retention_seconds only takes effect when offset_commit is v2-4, this worth documenting in changelog.

hint the users that starting from v5, the broker-side `offsets.retention.minutes` applies instead.

src/brod_group_coordinator.erl
```diff
@@ -909,13 +910,15 @@ do_commit_offsets_(#state{ groupId                  = GroupId
...
                          } = State) ->
+  StaticMemberID = request_group_instance_id(State),
   Metadata = make_offset_commit_metadata(),
   TopicOffsets0 =
     brod_utils:group_per_key(
       fun({{Topic, Partition}, Offset}) ->
         PartitionOffset =
           [ {partition_index, Partition}
           , {committed_offset, Offset + 1} %% +1 since roundrobin_v2 protocol
+          , {committed_leader_epoch, -1}
```
zmstone: please add a short comment on the magic number, e.g. `%% -1 = unknown epoch

test/brod_kafka_request_tests.erl
```diff
@@ -0,0 +1,49 @@
...
+    , {retention_time_ms, -1}
+    , {topics,
+       [[ {name, <<"topic">>}
+        , {partitions, [Partition]}
+        ]]}
+    ],
+  Req = brod_kafka_request:offset_commit(Vsn, Fields),
+  ?assertMatch(#kpro_req{api = offset_commit, vsn = Vsn}, Req),
+  Encoded = iolist_to_binary(kpro:encode_request(<<"brod">>, 1, Req)),
+  ?assert(is_binary(Encoded)).
```
zmstone: it's always true.
if the idea is to assert instance id is encoded, maybe match the binary with substring `<<"instance">>` will do

zmstone: even better if we use a unique enough string for `group_instance_id`

whatyouhide: Done. 👍

src/brod_group_coordinator.erl
whatyouhide: Done!

src/brod_group_coordinator.erl
whatyouhide: Done 👍

zmstone (approved):

PR #674, opened 2026-08-25
Make static-member fencing configurable

whatyouhide:
Static member fencing has two valid recovery policies:

* Retrying lets a replacement recover if the old process temporarily rejoins while it is shutting down.
* Stopping prevents two long-lived processes from repeatedly replacing and fencing each other.

This PR adds the `fenced_member_action` group option:

* `retry` keeps the existing behavior and remains the default.
* `stop` stops the group coordinator with reason `fenced_instance_id`.

The retry path keeps its existing minimum 10-second delay and uses the existing rejoin attempt handling. The stop path revokes local assignments, collects pending acknowledgements, and exits without trying to rejoin.

This keeps the current behavior for existing brod users while allowing applications with their own fenced-member handling to select terminal fencing.

The coordinator tests cover both the default retry policy and the opt-in stop policy.


src/brod_group_coordinator.erl
```diff
@@ -663,27 +661,15 @@ is_commit_possible(?unknown_member_id) -> false;
...
-schedule_retry_after_fencing(
-  #state{rejoin_delay_seconds = RejoinDelaySeconds} = State,
-  AttemptNo
-) ->
-  RetryDelaySeconds = fenced_retry_delay_seconds(RejoinDelaySeconds),
-  log(State, error,
-      "Static member was fenced by another member using the same "
-      "group_instance_id; check for duplicate group member configuration. "
-      "Retrying stabilization after ~p seconds.",
-      [RetryDelaySeconds]),
```
zmstone, line 679: this error level log should be restored before shutdown

zmstone (commented):
please add a new chaneglog entry.
document what it means for group subscriber supervisor.
because coordinator will exit `{stop, {shutdown, coordinator_failure}}`, a permanent strategy supervisor may:

- before hitting intensity, bring back the child, and run into the same mutual fencing loop
- hit max restart intensity, supervisor may restart, run into the same mutual fencing loop
- bring down the application ...

i.e. we can only hope the supervisor to use transient, or temporary, otherwise i am unsure if this actually will improve the situation.

src/brod_group_coordinator.erl
```diff
@@ -243,7 +246,14 @@
...
 %%  <li>`rejoin_delay_seconds' (optional, default = 1)
 %%
-%%      Delay in seconds before re-joining the group.</li>
+%%      Delay in seconds before re-joining the group. Fenced members wait at
+%%      least 10 seconds between attempts.</li>
+%%
+%%  <li>`fenced_member_action' (optional, default = `retry')
+%%
+%%      What to do when another member uses the same `group_instance_id'.
+%%      `retry' keeps trying to re-join, subject to `max_rejoin_attempts'.
```
zmstone: "subject to `max_rejoin_attempts`" overstates the cap. 

a new fencing event from a heartbeat or a commit starts again at attempt 0, and a successful takeover resets the cycle, so the retry loop does not terminate while two members share the instance id (as you noted yourself in the discussion). 

Suggest stating that directly, e.g.: "`retry' re-joins to take the slot back, waiting at least 10 seconds between attempts, for as long as the conflict persists; `max_rejoin_attempts' bounds only consecutive failed join attempts within one cycle."

src/brod_group_coordinator.erl
```diff
@@ -243,7 +246,14 @@
...
+%%      `stop' stops the gen_server with reason `fenced_instance_id'.</li>
```
zmstone: suggest documenting here what the stop means one level up, 
because that decides whether `stop' is actually terminal: `brod_group_subscriber_v2' turns any coordinator exit into 
`{stop, {shutdown, coordinator_failure}}`, 

so a `transient' subscriber stays down, while a `permanent' 
one restarts, re-joins with the same instance id, and resumes 
the fencing cycle at supervisor-restart cadence. 

`brod_group_subscriber` (v1) does not trap exits, so it exits with reason `fenced_instance_id' itself.

src/brod_group_coordinator.erl
```diff
@@ -472,34 +489,37 @@ handle_info({msg, _Pid, #kpro_rsp{ api = heartbeat
...
-    {ok, NewState_} = stabilize(State, 0, Reason),
-    {noreply, NewState_}
+  catch
+    throw : ?fenced_instance_id ->
+      gen_server:reply(From, {error, ?fenced_instance_id}),
+      handle_stabilize_result(
+        handle_fencing(StateWithOffsets, 0, preparation_required));
+    throw : Reason ->
+      gen_server:reply(From, {error, Reason}),
+      handle_stabilize_result(stabilize(State, 0, Reason))
```
zmstone: looks like a bug in old code, this catch clause should actually continue with `StateWithOffsets`.

src/brod_group_coordinator.erl
```diff
@@ -580,31 +600,38 @@ is_already_connected(#state{connection = Conn}, {Host, Port}) ->
...
+                         RejoinDelaySeconds).
+
+continue_stabilization({stop, ?fenced_instance_id, _State} = Stop,
+                       _AttemptNo, _Reason, _RejoinDelaySeconds) ->
+  Stop;
+continue_stabilization({ok, State2}, AttemptNo, Reason,
+                       RejoinDelaySeconds) ->
   State3 = State2#state{is_in_group = false},
 
   %$ 4. Clean up state based on the last failure reason
```
zmstone: Pre-existing typo on this moved line: `%$' should be `%%'.

whatyouhide: Ah nice, fixed.

src/brod_group_coordinator.erl
whatyouhide: Fixed.

src/brod_group_coordinator.erl
whatyouhide: I added a not about this + sup strategies, lmk how it's looking.

src/brod_group_coordinator.erl
whatyouhide: Nice, done.

zmstone (approved):

Issue #419, opened 2020-09-23
Producer 'function clause' error


pouriya:
After producing 31 millions data I see following errors for all producers in logs:  
```erlang
2020-09-23 11:19:01.968201+00:00 [error] ** Generic server <0.629.0> terminating 
** Last message in was {msg,<0.24609.0>,
                            {kpro_rsp,#Ref<0.3617632699.303562781.156674>,
                                      produce,5,
                                      #{responses =>
                                            [#{partition_responses =>
                                                   [#{base_offset => 177509,
                                                      error_code => no_error,
                                                      log_append_time => -1,
                                                      log_start_offset => 0,
                                                      partition => 106}],
                                               topic =>
                                                   <<"topic-name">>}],
                                        throttle_time_ms => 0}}}
** When Server state == {state,<0.481.0>,<<"topic-name">>,106,
                            <0.24609.0>,#Ref<0.3617632699.300154889.31499>,
                            {buf,1000,1,10485760,3,0,0,
                                #Fun<brod_producer.1.19472292>,37,0,
                                {[],[]},
                                {[{req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
                                  {req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
                                  {req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
                                  {req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
                                  {req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
                                  {req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
                                  {req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
                                  {req,#Fun<brod_producer.2.19472292>,
                                      #Fun<brod_producer_buffer.0.67628820>,
                                      680,1,1600859941968,0},
[...]

PR #675, opened 2026-09-04
Fix function_clause crash on stale produce responses

whatyouhide:
I think this was already described in #419 but we've seen it happen in production a few hundred times so far. We set `partition_onwire_limit` to `5`, for context.

cc @zmstone

### DIFF
--- a/src/brod_producer_buffer.erl
+++ b/src/brod_producer_buffer.erl
@@ -28,6 +28,7 @@
         ]).
 
 -export([ is_empty/1
+        , is_onwire/2
         ]).
 
 -export_type([buf/0]).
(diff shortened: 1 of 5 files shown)


whatyouhide:
Ping @zmstone 🙏

src/brod_producer_buffer.erl
zmstone: the cause could be a stale ack
to handle stale, we simply add a clause above the current:
`ack(#buf{onwire = []} = Buf, _, _) -> Buf;`

if the ack is out of order, i am unsure how to handle, 
because it breaks the ordering guarantee to caller(s).

